PR #7, opened 2020-02-21
0.2.9

monkey-jeff:


monkey-jeff:
Doing this because I rereleased a new version with a bugfix for Daniel Barbato.  Tagged the release in its own PR though.  (Already pushed the version to NPM)

PR #12, opened 2020-04-21
Update exchange-connector tests and package.json

gkhunger-sf:


test/exchange-connector/exchangeTools.test.ts
```diff
@@ -184,3 +186,43 @@ describe("Test groupByCategory method", () => {
...
+        createdDate: "today",
+        mainFile: "api.raml"
+      }
+    }
+  ];
+
+  it("removes fat raml external links for all the apis", () => {
+    const apis = _.cloneDeep(REST_APIS);
+    expect(removeRamlLinks(apis)).to.not.contain("externalLink");
+  });
```
kalagarraj, line 227: Can we pls add tests for arrays that are 
1. empty
2. null
3. undefined
4. containing objects without external Link

gkhunger-sf: done

src/exchange-connector/exchangeTools.ts
```diff
@@ -41,6 +41,9 @@ export function groupByCategory(
  * @returns {RestApi[]} Same list of APIs but without their external links
  */
 export function removeRamlLinks(apis: RestApi[]): RestApi[] {
+  if(!apis) {
```
wjhsf, line 44: Why is this needed? Type `RestApi[]` should never be a falsy value.

test/exchange-connector/exchangeTools.test.ts
```diff
@@ -184,3 +186,64 @@ describe("Test groupByCategory method", () => {
...
+    });
+
+    expect(removeRamlLinks(apis)).to.not.contain("externalLink");
+  });
+
+  it("should not do anything if an empty array is passed", () => {
+    expect(removeRamlLinks([])).to.deep.equal([]);
+  });
+
+  it("should return null if null is passed", () => {
```
wjhsf, line 242: Is this the desired behavior? Based on the function's usage, I would consider undefined/null to be unexpected input, and unexpected input should generally throw an error.

gkhunger-sf: that's a good point. For this story I have been not throwing errors because we don't want to halt the entire process in most cases but in this case it doesn't make sense to go forward if the input is null/undefined. I'll make the change.

test/exchange-connector/exchangeTools.test.ts
```diff
@@ -184,3 +186,64 @@ describe("Test groupByCategory method", () => {
...
+    expect(removeRamlLinks(apis)).to.not.contain("externalLink");
+  });
+
+  it("should not fail if the RAML does not have externalLink fields", () => {
+    const apis = _.cloneDeep(REST_APIS);
+    apis.forEach(apiEntry => {
+      delete apiEntry.fatRaml.externalLink;
+    });
+
+    expect(removeRamlLinks(apis)).to.not.contain("externalLink");
```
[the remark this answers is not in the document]
gkhunger-sf: I addressed it. Please take a look.

PR #14, opened 2020-04-22
 Skip and continue if downloading an api from exchange fails

gkhunger-sf:
These changes are to help with handle instances where exchange-connector fails to get information about and download an asset. 

- The entire download process doesn't halt if that happens
- Adds log messages with urls to go to manually get the resources
- Adds removeVersionSpecificInformation to remove all the information in api-config.json in the event of a failure
- Converts downloadRestApi and getAsset into async functions

### DIFF
--- a/src/exchange-connector/exchangeDownloader.ts
+++ b/src/exchange-connector/exchangeDownloader.ts
@@ -12,38 +12,32 @@ import fetch, { Response } from "node-fetch";
 import path from "path";
 
 import { RestApi, FileInfo, Categories } from "./exchangeTypes";
+import { ramlToolLogger } from "../logger";
 
 const DEFAULT_DOWNLOAD_FOLDER = "download";
 const ANYPOINT_BASE_URI = "https://anypoint.mulesoft.com/exchange/api/v2";
+const ANYPOINT_BASE_URI_WITHOUT_VERSION =
+  "https://anypoint.mulesoft.com/exchange";
 
-export function downloadRestApi(
+export async function downloadRestApi(
   restApi: RestApi,
   destinationFolder: string = DEFAULT_DOWNLOAD_FOLDER
 ): Promise<void | Response> {
-  return new Promise((resolve, reject) => {
-    if (!restApi.fatRaml) {
-      reject(
-        new Error(
-          `Fat RAML download information for ${restApi.assetId} is missing`
-        )
-      );
-    }
+  if (!restApi.id) {
+    ramlToolLogger.info(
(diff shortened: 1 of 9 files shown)


src/exchange-connector/exchangeDownloader.ts
```diff
@@ -12,38 +12,32 @@ import fetch, { Response } from "node-fetch";
...
-      reject(
-        new Error(
-          `Fat RAML download information for ${restApi.assetId} is missing`
-        )
-      );
-    }
-
-    ensureDirSync(destinationFolder);
+  if (!restApi.id) {
+    ramlToolLogger.info(
```
unandyala, line 27: should this be a warning?

unandyala: It might to be useful if the log message says RestApi ID is missing

src/exchange-connector/exchangeDownloader.ts
```diff
@@ -111,17 +105,24 @@ function convertResponseToRestApi(apiResponse: JSON): RestApi {
...
     }
-  }).then(res => {
-    if (!res.ok) {
-      throw new Error(`${res.status} - ${res.statusText}`);
-    }
-    return res.json();
   });
+  if (!res.ok) {
+    ramlToolLogger.info(
+      `Failed to get information about ${assetId} from exchange: ${res.status} - ${res.statusText}`,
```
unandyala, line 119: same, I feel this is a warning

src/exchange-connector/exchangeDownloader.ts
gkhunger-sf: Yeah these are warnings. The only reason I made them infos is because with our current config warnings are not displayed.

gkhunger-sf: We are only checking the ID but other information is missing as well.

src/exchange-connector/exchangeDownloader.ts
```diff
@@ -159,29 +160,33 @@ export function searchExchange(
  * @param {RegExp} deployment
  * @returns {Promise<string>} Returned the version string that matches the regex passed.  Will return first found result
  */
-export function getVersionByDeployment(
+export async function getVersionByDeployment(
```
bredmond-sf, line 163: [PRAISE] thank you for refactoring! much easier to read

gkhunger-sf: 🙏

PR #19, opened 2020-04-24
Update mercury profile to use the new AMF vocabulary

gkhunger-sf:
- Update the amf-client version to 4.0.4
- Update mercury profile to use the new AMF vocabulary
- Replace npm install with npm ci
- Change copyright year to 2020

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
       - checkout
       - node/with-cache:
           steps:
-            - run: npm install
+            - run: npm ci
       - run: npm run test
       - store_test_results:
           path: reports
--- a/test/utils.test.ts
+++ b/test/utils.test.ts
@@ -44,10 +44,8 @@ export function breaksOnlyOneRule(
   rule: string
 ): void {
   assert.equal(result.conforms, false, result.toString());
-  if (!result.conforms) {
-    assert.equal(result.results.length, 1, result.toString());
-    assert.equal(result.results[0].validationId, rule, result.toString());
-  }
+  assert.equal(result.results.length, 1, result.toString());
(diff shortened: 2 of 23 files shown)


test/utils.test.ts
```diff
@@ -44,8 +44,10 @@ export function breaksOnlyOneRule(
   rule: string
 ): void {
   assert.equal(result.conforms, false, result.toString());
-  assert.equal(result.results.length, 1, result.toString());
-  assert.equal(result.results[0].validationId, rule, result.toString());
+  if (!result.conforms) {
```
kalagarraj, line 47: Why do we need to check for conforms? is it throwing any error with the new versioin
Do we have the following tests 

1. undefined result.confirms 
2. null result.conforms

gkhunger-sf: doesn't seem to be failing without the check. So I will revert the change.

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       - checkout
       - node/with-cache:
           steps:
-            - run: npm install
+            - run: npm ci
```
monkey-jeff, line 10: What does `npm ci` do?

gkhunger-sf: In short, the main differences between using npm install and npm ci are:

- The project must have an existing package-lock.json or npm-shrinkwrap.json.
- If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.
- npm ci can only install entire projects at a time: individual dependencies cannot be added with this command.
- If a node_modules is already present, it will be automatically removed before npm ci begins its install.
- It will never write to package.json or any of the package-locks: installs are essentially frozen.

PR #20, opened 2020-04-27
Br/baseuriparameter only shortcode.w 7275821

bredmond-sf:
baseUriParameter section *MUST* only have  shortCode rule for ramllint.

To run tests
```bash
TS_NODE_COMPILER_OPTIONS='{"rootDir":"."}' npx mocha test/mercury/baseuriparameters.test.ts
```

### DIFF
--- a/test/mercury/baseuriparameters.test.ts
+++ b/test/mercury/baseuriparameters.test.ts
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+/* eslint-disable no-undef */
+"use strict";
+import { validateFile } from "../../src/validator";
+import {
+  getHappySpec,
+  conforms,
+  breaksOnlyOneRule,
+  renderSpecAsFile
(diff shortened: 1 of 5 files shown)


test/mercury/baseuriparameters.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    // We have to change the baseUri because it's implicit if it appears in the template
+    doc["baseUri"] = "https://example.com";
+    delete doc["baseUriParameters"]["shortCode"];
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+    breaksOnlyOneRule(
+      result,
+      "http://a.ml/vocabularies/data#baseuriparameters-must-only-have-shortcode"
+    );
+  });
+
```
kalagarraj, line 37: Pls add a test for url like "https://{shortCode}.api.commercecloud.salesforce.com/{anotherParameterInPlaceOfApiFamily}/shopper-orders/{version}/";

bredmond-sf: done

bredmond-sf:
This is covered as much as possible by existing rules. We have a rule enforcing the correct baseUri pattern which enforces only having shortCode (and version, which is special). If someone specifies something in baseUriParameters that's not in the uri, that is invalid RAML. The one thing we can't check for sure is that shortCode is in baseUriParameters explicitly, but it's technically optional anyway.

PR #21, opened 2020-04-28
Add a rule to check baseUri pattern

gkhunger-sf:
There should be a base uri and it should match this pattern:
https://{shortCode}.api.commercecloud.salesforce.com/test-family/test-api/{version}

### DIFF
--- a/profiles/mercury.raml
+++ b/profiles/mercury.raml
@@ -27,6 +27,7 @@ violation:
   - require-json
   - resource-name-validation
   - version-format
+  - base-uri-matches-pattern
 
 validations:
 
@@ -159,4 +160,14 @@ validations:
         pattern: ^v[0-9]+$
         minCount: 1
 
+  base-uri-matches-pattern:
+    message: The baseUri must match the pattern - https://{shortCode}.api.commercecloud.salesforce.com/<api-family>/<api-name>/{version}
+    targetClass: apiContract.WebAPI
+    functionConstraint:
+      code: |
+        function(resource) {
+          const baseUriRegex = /^https:\/\/{shortCode}.api.commercecloud.salesforce.com\/[a-z]+-?[a-z]+\/[a-z]+-?[a-z]+\/{version}$/;
+          const server = resource['apiContract:server'];
+          return server ? baseUriRegex.test(server[0]['core:urlTemplate'][0]) : false;
+        }
 
--- a/test/mercury/base-uri-pattern.test.ts
+++ b/test/mercury/base-uri-pattern.test.ts
@@ -1,0 +1,136 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+/* eslint-disable no-undef */
+"use strict";
+import {
+  getHappySpec,
+  renderSpecAsFile,
+  breaksOnlyOneRule
+} from "../utils.test";
+import { validateFile } from "../../src/validator";
+
+const PROFILE = "mercury";
+const RULE = "http://a.ml/vocabularies/data#base-uri-matches-pattern";
+
+describe("base uri pattern validation", () => {
+  let doc;
+
+  beforeEach(() => {
+    doc = getHappySpec();
+  });
+
+  it("should not conform if the baseUri is missing", async () => {
+    delete doc.baseUri;
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    breaksOnlyOneRule(result, RULE);
+  });
+
+  it("should not conform if the protocol is http", async () => {
+    doc["baseUri"] =
+      "http://{shortCode}.api.commercecloud.salesforce.com/test-family/test-api/{version}";
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    breaksOnlyOneRule(result, RULE);
+  });
+
+  it("should not conform if the protocol is missing", async () => {
+    doc["baseUri"] =
+      "{shortCode}.api.commercecloud.salesforce.com/test-family/test-api/{version}";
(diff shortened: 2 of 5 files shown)


test/mercury/base-uri-pattern.test.ts
```diff
@@ -0,0 +1,136 @@
...
+const PROFILE = "mercury";
+const RULE = "http://a.ml/vocabularies/data#base-uri-matches-pattern";
+
+describe("base uri pattern validation", () => {
+  let doc;
+
+  beforeEach(() => {
+    doc = getHappySpec();
+  });
+
```
unandyala, line 25: add test for happy path?

test/mercury/base-uri-pattern.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    breaksOnlyOneRule(result, RULE);
+  });
+
+  it("should not conform if the baseUri doesn't end with {version}", async () => {
+    doc["baseUri"] =
+      "https://{shortCode}.api.commercecloud.salesforce.com/test-family/test-api/{version}/";
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    breaksOnlyOneRule(result, RULE);
+  });
```
kalagarraj, line 135: Nice number of tests for not conforming..

Can u pls add one for conforming the shortCode rule?

test/mercury/base-uri-pattern.test.ts
gkhunger-sf: done

test/mercury/base-uri-pattern.test.ts
gkhunger-sf: Added a happy path test

profiles/mercury.raml
```diff
@@ -159,4 +160,14 @@ validations:
         pattern: ^v[0-9]+$
         minCount: 1
 
+  base-uri-matches-pattern:
+    message: The baseUri must match the pattern - https://{shortCode}.api.commercecloud.salesforce.com/<api-family>/<api-name>/{version}
+    targetClass: apiContract.WebAPI
+    functionConstraint:
+      code: |
+        function(resource) {
```
monkey-jeff, line 168: Why code here instead of a pattern regex enforcement?

profiles/mercury.raml
```diff
@@ -159,4 +160,14 @@ validations:
...
+          const baseUriRegex = /^https:\/\/{shortCode}.api.commercecloud.salesforce.com\/[a-z]+-?[a-z]+\/[a-z]+-?[a-z]+\/{version}$/;
+          const server = resource['apiContract:server'];
+          return server ? baseUriRegex.test(server[0]['core:urlTemplate'][0]) : false;
```
monkey-jeff, line 171: server[0] seems a bit easy to break.  Can you have multiple servers specified?  Not sure we want to do this with custom code.

profiles/mercury.raml
```diff
@@ -159,4 +160,14 @@ validations:
         pattern: ^v[0-9]+$
         minCount: 1
 
+  base-uri-matches-pattern:
+    message: The baseUri must match the pattern - https://{shortCode}.api.commercecloud.salesforce.com/<api-family>/<api-name>/{version}
+    targetClass: apiContract.WebAPI
+    functionConstraint:
+      code: |
+        function(resource) {
+          const baseUriRegex = /^https:\/\/{shortCode}.api.commercecloud.salesforce.com\/[a-z]+-?[a-z]+\/[a-z]+-?[a-z]+\/{version}$/;
```
monkey-jeff, line 169: This does not allow a trailing / in the regex.  Is this intended?

test/mercury/base-uri-pattern.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      "https://{shortCode}.api.commercecloud.salesforce.com/test-family/test-api";
+
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    breaksOnlyOneRule(result, RULE);
+  });
+
+  it("should not conform if the baseUri doesn't end with {version}", async () => {
+    doc["baseUri"] =
+      "https://{shortCode}.api.commercecloud.salesforce.com/test-family/test-api/{version}/";
```
monkey-jeff, line 151: I think this should probably be a success.

test/mercury/base-uri-pattern.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  it("should not conform if the api family or the api name is missing", async () => {
+    doc["baseUri"] =
+      "https://{shortCode}.api.commercecloud.salesforce.com/test-api/{version}";
+
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    breaksOnlyOneRule(result, RULE);
+  });
+
+  it("should not conform if the version is missing", async () => {
```
monkey-jeff, line 140: Should have tests with something like `{VERSION}` as well

test/mercury/base-uri-pattern.test.ts
```diff
@@ -0,0 +1,157 @@
+/*
```
monkey-jeff, line 1: SHould try breaking your code some more here.  Do things like double `/`between segments.

profiles/mercury.raml
gkhunger-sf: To validate that a baseUri is part of the spec. `apiContract.urlTempalte` is nested inside `apiContract.Server`. When baseUri is not present in the raml, it doesn't even execute propertyConstraints.

profiles/mercury.raml
gkhunger-sf: Yes, it is intended. If we allow a trailing / we will end up with // when the baseUri is appended with endpoints.

monkey-jeff: Followup...does anyone currently have a trailing / in their base uri.  If not this is fine.  If so I don't think we should force people to change it right now this close to GA.

gkhunger-sf: Yeah, shopper-stores and shopper-search. I will remove the check.

PR #22, opened 2020-04-28
Update the base uri rule

gkhunger-sf:
* Allow a forward slash at the end of the base uri
* Fix the regex so that the rule doesn't fail if the api family and the api name have more than two words
* Add more tests

### DIFF
--- a/profiles/mercury.raml
+++ b/profiles/mercury.raml
@@ -166,7 +166,7 @@ validations:
     functionConstraint:
       code: |
         function(resource) {
-          const baseUriRegex = /^https:\/\/{shortCode}.api.commercecloud.salesforce.com\/[a-z]+-?[a-z]+\/[a-z]+-?[a-z]+\/{version}$/;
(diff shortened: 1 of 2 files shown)


profiles/mercury.raml
```diff
@@ -167,7 +167,19 @@ validations:
...
         function(resource) {
           const baseUriRegex = /^https:\/\/{shortCode}.api.commercecloud.salesforce.com\/([a-z]+-?)+\/([a-z]+-?)+\/{version}\/?$/;
-          const server = resource['apiContract:server'];
-          return server ? baseUriRegex.test(server[0]['core:urlTemplate'][0]) : false;
+          const servers = resource['apiContract:server'];
+          if(servers.length < 1) {
+            return false;
+          }
+          let result = true;
+          servers.forEach(server => {
```
unandyala, line 175: Is there a possibility of more than one server?

gkhunger-sf: I don't think so. Made this change just to make Jeff happy 😛

unandyala: can I suggest to check for >1 as well. probably this is a separate rule that executes before this one

gkhunger-sf: I like the idea of a separate rule for this. I will implement it in the next PR.

PR #32, opened 2020-05-13
[jb/validator-print-tests.W-7559825]

jbegleiter-sf:
- Ordered package.json dep list
- Updated utils.test -> testUtils to avoid confusion
- Added validator.test.ts specifically to test print functionality
- Updated branch logic for validator.ts to avoid nesting / reduce branches


test/validator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  afterEach(() => {
+    consoleSpy.restore();
+  });
+
+  it("doesn't console log when no results passed", async () => {
+    await printResults(null);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
```
monkey-jeff, line 39: Ummmm....what is the error that this was producing and why are we ignoring it?

test/validator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    return expect(consoleSpy.callCount).to.equal(0);
+  });
+
+  it("outputs results in first branch when results conform and warnings are false (default)", async () => {
+    const doc = getHappySpec();
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    await printResults(result);
```
monkey-jeff, line 47: WOAH! Josh using an await?

test/validator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
+    return expect(consoleSpy.getCall(0).args[0]).to.have.string(
```
monkey-jeff, line 51: What are we testing here?  There is no validation that the above call does not have the warnings.  Basically, if warnings are false or true won't this pass?

test/validator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  it("outputs results in second branch when results do not conform and warnings are false (default)", async () => {
+    const doc = getSingleInvalidFile();
+    const result = await validateFile(renderSpecAsFile(doc), PROFILE);
+
+    await printResults(result);
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
+    return expect(consoleSpy.getCall(0).args[0]).to.have.string(
```
monkey-jeff, line 64: Same as above.  To elaborate the `console.log(results.toString());` will produce the message of `Conforms? false|true` just like our print does.

test/validator.test.ts
jbegleiter-sf: This actually isn't a problem, I resolved it by using this semantic instead of `to.have.been.called`, and then didn't remove this, and then copy pasta. I'll remove.

test/validator.test.ts
jbegleiter-sf: Yes, I actually need to ensure the args DOESN'T include something that would only be included in the second branch, good cal out

test/validator.test.ts
jbegleiter-sf: No, in this case because warnings are false the only way `Conforms? false` can appear in the console log is if the results have conforms false, meaning we are in the second branch

monkey-jeff: Unless our if logic is wrong.  All I am saying is that regardless of the conditional this line as is CAN appear in either branch.  

If we make the assumption that our conditional logic is correct then why test the branches?

If someone makes a change in the future to the conditional logic that breaks it this should cause a failure and I'm not sure it will here. 

I won't block it on these grounds but would love to see this updated.

jbegleiter-sf: Updated to be more specific

PR #37, opened 2020-05-13
Add tests for exchangeDirectoryParser

gkhunger-sf:
Brings the overage to 100%

### DIFF
--- a/src/exchange-connector/exchangeDirectoryParser.ts
+++ b/src/exchange-connector/exchangeDirectoryParser.ts
@@ -25,7 +25,7 @@ function getFiles(directory): fs.Dirent[] {
 export function extractFiles(
   directory: string,
   removeFiles = true
-): Promise<void> {
+): Promise<void[]> {
   const files: fs.Dirent[] = getFiles(directory);
   const promises: Promise<void>[] = [];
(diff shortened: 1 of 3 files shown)


src/exchange-connector/exchangeDirectoryParser.ts
```diff
@@ -52,8 +53,7 @@ export function extractFiles(
           );
         })
       );
-    }
-  });
+    });
 
   // The then here collapses the return from an array of void to a single void
```
wjhsf, line 58: Minor issue, but [this is not true](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then).

src/exchange-connector/exchangeDirectoryParser.ts
```diff
@@ -52,9 +53,7 @@ export function extractFiles(
           );
         })
       );
-    }
-  });
+    });
 
-  // The then here collapses the return from an array of void to a single void
   return Promise.all(promises).then();
```
wjhsf, line 58: `.then()` here is unnecessary - it just clones the promise, which we don't need to do.

PR #46, opened 2020-05-27
Consolidate RAML to AMF parser code

bredmond-sf:
We should NOT have 2 methods with different functionality that parses raml files.

As a steelarc dev I want to use the parseRaml method from raml-toolkit so that I can benefit from its advanced capabilities (e.g. support for nested dependencies).

Currently, the SDK generator has its own implementation of "parseRamlFile" method while the raml-toolkit has a far superior "parseRaml" method with better support for nested dependencies. We should be using that in the commerce SDK and deprecate the parseRamlFile method.

This integrates parseRamlFile into raml-toolkit. Another PR will remove it from the SDK.


src/parser.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Extract all of the delcared data types from an AMF model.
+ *
+ * @param api - The model to extract data types from
+ *
+ * @returns data types from model
+ */
+export function getAllDataTypes(
+  api: model.document.BaseUnit & amf.model.document.DeclaresModel
```
monkey-jeff, line 109: This might be my fault, but why `model.document..` and `amf.model.document...`

package.json
```diff
@@ -148,7 +148,8 @@
     "nyc": "^15.0.0",
     "prettier": "^1.15.2",
     "sinon": "^9.0.2",
-    "tslib": "^1.10.0"
+    "tslib": "^1.10.0",
+    "webapi-parser": "^0.4.1"
```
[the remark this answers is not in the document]
bredmond-sf: Hope so because it's in the SDK :D

src/parser.ts
```diff
@@ -0,0 +1,177 @@
+/*
```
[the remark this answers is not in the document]
bredmond-sf: parser.ts is almost exactly what was in SDK, except we're using the parseRamlFile from raml-toolkit. Otherwise, pretty much the same.

src/parser.ts
bredmond-sf: Fixed

src/parser.ts
```diff
@@ -0,0 +1,176 @@
...
+  const fatRamlResourceLoader = new FatRamlResourceLoader(
+    path.dirname(fileUri)
+  );
+  const ccEnvironment = amf.client.DefaultEnvironment.apply().addClientLoader(
+    fatRamlResourceLoader
+  );
+  const parser = new amf.Raml10Parser(ccEnvironment);
+
+  let model: model.document.BaseUnit;
+  try {
```
unandyala, line 43: I know this is copied from existing code but thought we can make changes while we are here -
Why is there a try-catch around this specific call?  Should we just throw the error since this is now a common function.
If not throwing the error then just change the error message to be more specific and use the logger instead of console.log

test/parser.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  WebApiBaseUnitWithEncodesModel
+} from "webapi-parser";
+
+import { expect, default as chai } from "chai";
+import path from "path";
+import chaiAsPromised from "chai-as-promised";
+import _ from "lodash";
+
+before(() => {
+  chai.use(chaiAsPromised);
```
wjhsf, line 29: chai-as-promised is never actually used. Tests involving promises should either be cleaned up and use it, or they should be converted to async functions and chai-as-promised can be removed.

test/parser.test.ts
```diff
@@ -0,0 +1,197 @@
...
+        "result_page"
+      ]);
+    });
+  });
+
+  it("Test valid RAML file with references", () => {
+    const ramlFile = path.join(__dirname, "site.raml");
+    parseRamlFile(ramlFile)
+      .then(refModel => {
+        return parseRamlFile(ramlFile).then(mainModel => {
```
wjhsf, line 73: Nested promise chain inside a promise chain? :-1:

test/parser.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    );
+  });
+  it("Test with valid model and resolution pipeline", () => {
+    const ramlFile = path.join(__dirname, "site.raml");
+    return parseRamlFile(ramlFile)
+      .then(s => {
+        return resolveApiModel(s as WebApiBaseUnitWithEncodesModel, "editing");
+      })
+      .then(resolvedModel => {
+        expect(resolvedModel).to.exist;
```
wjhsf, line 194: Avoiding this is what chai-as-promised was designed for.

test/parser.test.ts
bredmond-sf: done

src/parser.ts
bredmond-sf: done

PR #47, opened 2020-05-28
Provide error message when profile fails to parse

monkey-jeff:
Need to write a test for this, but basically someone writing profiles without this they would always get the following error regardless.

```
➜  raml-toolkit git:(master) ✗ ./bin/run -p aip test/mercury.raml 
(node:84603) UnhandledPromiseRejectionWarning
(node:84603) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:84603) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

The error now looks like
```
➜  raml-toolkit git:(master) ✗ ./bin/run -p aip test/mercury.raml
Error parsing validation profile.
  Message: Syntax error in the following text: ' JSON SHOULD be used as the representation type (TODO: dupe/conflict?)'
  Target: 
Property: 
  Position: Some(LexicalInformation([(45,12)-(45,82)]))
 at location: Some(file:///Users/jalbert/src/raml-toolkit/profiles/aip.raml)
```

### DIFF
--- a/src/validator.ts
+++ b/src/validator.ts
@@ -19,7 +19,6 @@ export async function validateCustom(
     // We rethrow to provide a cleaner error message
     throw new Error(err.vw);
   }
-  console.log("SHOULD NOT HIT");
   const report = await amf.Core.validate(
     model,
     profileName,
--- a/src/validator.ts
+++ b/src/validator.ts
@@ -13,8 +13,14 @@ function validateCustom(
   profileFile: string
 ): Promise<amf.client.validate.ValidationReport> {
   return new Promise(async resolve => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const profileName: any = await amf.Core.loadValidationProfile(profileFile);
+    let profileName: amf.ProfileName;
+    try {
+      profileName = await amf.Core.loadValidationProfile(profileFile);
+    } catch (err) {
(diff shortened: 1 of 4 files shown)


src/validator.ts
```diff
@@ -13,8 +13,14 @@ function validateCustom(
   profileFile: string
 ): Promise<amf.client.validate.ValidationReport> {
   return new Promise(async resolve => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const profileName: any = await amf.Core.loadValidationProfile(profileFile);
+    let profileName: amf.ProfileName;
+    try {
```
kalagarraj, line 17: Are you working on tests?

src/validator.ts
```diff
@@ -13,8 +13,14 @@ function validateCustom(
...
+      profileName = await amf.Core.loadValidationProfile(profileFile);
+    } catch (err) {
+      console.error("Error parsing validation profile.");
```
unandyala, line 20: use logger?

monkey-jeff: Not needed anymore!

PR #49, opened 2020-06-01
Initial implementation to apply rules on differences

unandyala:
- Integrated [json-rules-engine](https://github.com/CacheControl/json-rules-engine) to apply rules on the differences
- Implemented display name change rule

**Changes to the difference object:**

- Added type property. This is the node type that has AMF vocabulary
- Added rule details. When the rule is evaluated to true, rule details are saved

**Example - Diff object with rule details**:

```json
{
    "id": "#/web-api/end-points/%2Forganizations%2F%7BorganizationId%7D%2Fproducts%2F%7Bid%7D/get",
    "type": [
      "apiContract:Operation",
      "doc:DomainElement"
    ],
    "added": {
      "core:name": "getProductDetails"
    },
    "removed": {
      "core:name": "getProduct"
    },
    "rule": {
      "name": "Rule to detect display name changes",
      "type": "display-name-change",
      "params": {
        "category": "Breaking"
      }
    }
  },
```


src/diffTool/rulesProcessor.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/**
+ * Get rules from rules json file
+ * @param rulesPath Path to rules json file
+ *
+ * @returns Array of rules
+ */
+function getRules(rulesPath: string): RuleProperties[] {
+  if (rulesPath == null) {
```
bredmond-sf, line 61: Isn't this also invalid? Why not throw here too?

src/diffTool/rulesProcessor.ts
```diff
@@ -0,0 +1,131 @@
...
+  let rules: RuleProperties[];
+  try {
+    rules = JSON.parse(fs.readFileSync(rulesPath, "UTF-8"));
+  } catch (error) {
+    error.message = `Error parsing the rules file '${rulesPath}': ${error.message}`;
+    throw error;
+  }
+  if (!Array.isArray(rules)) {
+    throw new Error(
+      `Rules must be defined as as a json array: ${JSON.stringify(
```
bredmond-sf, line 76: "as as"

src/diffTool/jsonDiff.ts
```diff
@@ -40,7 +41,21 @@ export enum DiffType {
 export class NodeDiff {
   public added?: { [key: string]: any };
   public removed?: { [key: string]: any };
-  constructor(public id: string) {
+  public rule?: {
+    //name of the rule that is evaluated to true
```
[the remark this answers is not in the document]
unandyala: sure. I will add

src/diffTool/diffProcessor.ts
```diff
@@ -29,7 +32,11 @@ export async function diffRaml(
   ramlToolLogger.info(
     `Finding differences between flattened JSON-LD of ${leftRaml} and ${rightRaml}`
   );
-  return findJsonDiffs(leftGraph, rightGraph);
+  const diffs = findJsonDiffs(leftGraph, rightGraph);
+  if (rules != null) {
```
kalagarraj, line 36: Do we need to check for non empty rules before applying?

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  const noRulesMsg = "No rules to apply on the differences";
+  const diffs = [new NodeDiff("test")];
+  it("exits when the rules file path is undefined", async () => {
+    await applyRules(diffs, undefined);
+    //verify that diff is not modified
+    expect(diffs).to.deep.equal(diffs);
+    //verify log message
+    sinon.assert.calledWith(loggerSpy, noRulesMsg);
+  });
+  it("exits when the rules file path is null", async () => {
```
kalagarraj, line 63: Add tests for empty rules file path

src/diffTool/jsonDiff.ts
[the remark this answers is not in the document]
bredmond-sf: nevermind. tsdoc really doesn't account for class properties

src/diffTool/diffProcessor.ts
unandyala: Its done in applyRules function

kalagarraj: Then I think we can do the null check in applyRules function as well

unandyala: null check is also done in applyRules, this is an additional check to avoid invoking if user do not want to apply rules

unandyala: null check is also done in applyRules. I will remove it here

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      id: "#/web-api/end-points/resource/get",
+      type: ["apiContract:Operation"],
+      added: { "core:name": "newName" },
+      removed: { "core:name": "oldName" }
+    };
+    await applyRules([diff], rulesPath);
+    const diffRule = diff.rule;
+    expect(diffRule.name).to.equal("Rule to detect display name changes");
+    expect(diffRule.type).to.equal("display-name-change");
+    expect(diffRule.params["category"]).to.equal("Breaking");
```
wjhsf, line 127: The diff rule has a known structure. Use `expect(diffRule).to.deep.equal(expectedStructure)`.

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+after(() => {
+  sinon.reset();
+});
+
+describe("Test that rules application exits with relevant message when there are no differences", () => {
+  const noDiffsMsg = "No differences to apply the rules";
+  it("exits when the differences are undefined", async () => {
+    const diffs = undefined;
+    await applyRules(undefined, rulesPath);
+    expect(diffs).to.equal(diffs);
```
wjhsf, line 36:
```suggestion
    await applyRules(undefined, rulesPath);
```
`undefined` is a primitive value, so `const diffs = undefined` can never be modified.

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  it("exits when there are empty differences", async () => {
+    const diffs = [];
+    await applyRules(diffs, rulesPath);
+    expect(diffs).to.deep.equal(diffs);
+    sinon.assert.calledWith(loggerSpy, noDiffsMsg);
+  });
+  it("exits when the differences are null", async () => {
+    const diffs = null;
+    await applyRules(null, rulesPath);
+    expect(diffs).to.equal(diffs);
```
wjhsf, line 48:
```suggestion
    await applyRules(null, rulesPath);
```
Same as above comment re: `undefined`.

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  it("exits when the differences are undefined", async () => {
+    const diffs = undefined;
+    await applyRules(undefined, rulesPath);
+    expect(diffs).to.equal(diffs);
+    sinon.assert.calledWith(loggerSpy, noDiffsMsg);
+  });
+  it("exits when there are empty differences", async () => {
+    const diffs = [];
+    await applyRules(diffs, rulesPath);
+    expect(diffs).to.deep.equal(diffs);
```
wjhsf, line 42:
```suggestion
    expect(diffs).to.deep.equal([]);
```
An object will always deep equal itself, so you need a second object to compare.

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+before(() => {
+  chai.should();
+  chai.use(chaiAsPromised);
+  loggerSpy = sinon.spy(ramlToolLogger, "info");
+});
+after(() => {
+  sinon.reset();
+});
+
+describe("Test that rules application exits with relevant message when there are no differences", () => {
```
wjhsf, line 31: The application isn't exiting in these tests, a function is returning a value and a message is being logged.

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  });
+  it("exits when the differences are null", async () => {
+    const diffs = null;
+    await applyRules(null, rulesPath);
+    expect(diffs).to.equal(diffs);
+    sinon.assert.calledWith(loggerSpy, noDiffsMsg);
+  });
+});
+
+describe("Test that rules application exits with relevant message when there are no rules", () => {
```
wjhsf, line 53: Same comment as the other `describe`.

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    await applyRules(diffs, tmpFile.name);
+    //verify that diff is not modified
+    expect(diffs).to.deep.equal(diffs);
+    //verify log message
+    sinon.assert.calledWith(loggerSpy, "No rules to apply on the differences");
+  });
+});
+
+describe("Test display name change rule ", () => {
+  const rulesPath = path.join(__dirname, "..", "..", "diffRules", "rules.json");
```
wjhsf, line 115:
```suggestion
  const rulesPath = path.join(__dirname, "../../diffRules/rules.json");
```

test/differencer/diffProcessor.test.ts
```diff
@@ -7,15 +7,73 @@
...
+    const rule = buildRule();
+    const tmpFile = tmp.fileSync({ postfix: ".json" });
+    fs.writeFileSync(tmpFile.name, `[${rule.toJSON()}]`);
+
+    const diffs = await diffRaml(leftRaml, rightRaml, tmpFile.name);
+    expect(diffs.length).to.greaterThan(0);
+    const diffRule = diffs[0].rule;
+    expect(diffRule.name).to.equal(rule.name);
+    expect(diffRule.type).to.equal(rule.event.type);
+    expect(diffRule.params).to.deep.equal(rule.event.params);
```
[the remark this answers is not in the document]
unandyala: for that I need to clone the original diff and add the expected details. I feel this is simple and easy to read as the test is mainly to validate the rule that is applied

test/differencer/rulesProcessor.test.ts
unandyala: it is the applyRules function that exits not the application. I will change the wording

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,129 @@
...
+/* eslint-disable @typescript-eslint/no-use-before-define*/
+
+const expect = chai.expect;
+const rulesPath = path.join(__dirname, "..", "..", "diffRules", "rules.json");
+let loggerSpy;
+before(() => {
+  chai.should();
+  chai.use(chaiAsPromised);
+  loggerSpy = sinon.spy(ramlToolLogger, "info");
+});
```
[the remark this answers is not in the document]
unandyala: top-level before make sense it you have something to clean-up in after.  In this case we can move chai calls out of before.

test/differencer/rulesProcessor.test.ts
unandyala: good catch

test/differencer/rulesProcessor.test.ts
unandyala: see the comment above

diffRules/rules.json
```diff
@@ -0,0 +1,33 @@
+[
```
monkey-jeff, line 1: What is this file? is it a default of some sort? or example? and should we name it appropriately to convey that?

test/differencer/rulesProcessor.test.ts
```diff
@@ -17,40 +17,36 @@ import chaiAsPromised from "chai-as-promised";
...
-  chai.should();
-  chai.use(chaiAsPromised);
   loggerSpy = sinon.spy(ramlToolLogger, "info");
 });
 after(() => {
   sinon.reset();
 });
 
-describe("Test that rules application exits with relevant message when there are no differences", () => {
+describe("Test that rules application function returns and logs a relevant message when there are no differences", () => {
```
wjhsf, line 31:
```suggestion
describe("Rules validator when RAML files have no differences", () => {
```

test/differencer/rulesProcessor.test.ts
```diff
@@ -17,40 +17,36 @@ import chaiAsPromised from "chai-as-promised";
...
-    const diffs = null;
+  it("logs message when the differences are null", async () => {
     await applyRules(null, rulesPath);
-    expect(diffs).to.equal(diffs);
     sinon.assert.calledWith(loggerSpy, noDiffsMsg);
   });
 });
 
-describe("Test that rules application exits with relevant message when there are no rules", () => {
+describe("Test that rules application function and logs a relevant message when there are no rules", () => {
```
wjhsf, line 49:
```suggestion
describe("Rules validator when rules file is invalid", () => {
```

src/diffTool/rulesProcessor.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/**
+ * Get rules from rules json file
+ * @param rulesPath - Path to rules json file
+ *
+ * @returns Array of rules
+ */
+function getRules(rulesPath: string): RuleProperties[] {
+  if (!rulesPath) {
```
monkey-jeff, line 61: This is only testing that the rulesPath is defined, not that it exists. Should we be doing that? And if not why don't we  let readJSONSync throw if rulesPath is undefined?

src/diffTool/rulesProcessor.ts
```diff
@@ -0,0 +1,122 @@
...
+async function successHandler(
+  event: Event,
+  almanac: Almanac,
+  ruleResult: RuleResult
+): Promise<void> {
+  const nodeDiff = (await almanac.factValue(DIFF_FACT_ID)) as NodeDiff;
+  ramlToolLogger.debug(
+    `Rule '${ruleResult.name}' is passed on difference '${nodeDiff.id}'`
+  );
+  nodeDiff.rule = {
```
monkey-jeff, line 92: What is this assignment doing? It is at the end of the method and I don't see how it is being used.

src/diffTool/rulesProcessor.ts
```diff
@@ -63,24 +63,22 @@ function getRules(rulesPath: string): RuleProperties[] {
   }
   let rules: RuleProperties[];
   try {
-    rules = JSON.parse(fs.readFileSync(rulesPath, "UTF-8"));
+    rules = fs.readJSONSync(rulesPath);
   } catch (error) {
     error.message = `Error parsing the rules file '${rulesPath}': ${error.message}`;
     throw error;
   }
```
wjhsf:
```suggestion
const rules: RuleProperties[] = fs.readJSONSync(rulesPath);
```
The modified error message doesn't provide any new information.

unandyala: readJSONSync does not provide user friendly message in all the cases like empty string

diffRules/rules.json
unandyala: This file has the rules defined that runs with update APIs.

src/diffTool/rulesProcessor.ts
```diff
@@ -0,0 +1,122 @@
...
+  await Promise.all(promises);
+}
+
+/**
+ * Get rules from rules json file
+ * @param rulesPath - Path to rules json file
+ *
+ * @returns Array of rules
+ */
+function getRules(rulesPath: string): RuleProperties[] {
```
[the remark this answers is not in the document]
unandyala: agree

diffRules/rules.json
monkey-jeff: I think default might be a better name for it then.  Or mercury.  Just trying to distinguish it more than calling it "rules"

src/diffTool/rulesProcessor.ts
unandyala: readJSONSync throws error when the file do not exist. I have a test for that
`/tmp/no-rules.json: ENOENT: no such file or directory`

src/diffTool/rulesProcessor.ts
unandyala: Adds rule that is passed/evaluated to true to the diff object

src/diffTool/rulesProcessor.ts
monkey-jeff: If that is the case, what purpose does this conditional server?

unandyala: Initially I was not modifying the error message in the catch and the default message was not user friendly. We don't need it anymore. I will remove it

diffRules/rules.json
unandyala: I will rename to **defaultRules.json**

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,126 @@
...
+import sinon from "sinon";
+import { ramlToolLogger } from "../../src";
+import chaiAsPromised from "chai-as-promised";
+
+/* eslint-disable @typescript-eslint/no-use-before-define*/
+
+const expect = chai.expect;
+chai.use(chaiAsPromised);
+
+const rulesPath = path.join(__dirname, "../../diffRules", "rules.json");
```
wjhsf, line 22: 1) Shouldn't this be `defaultRules.json`?
2) It looks like this is only used in the first `describe` block. Please move it there to avoid naming conflicts/confusion in other blocks.

unandyala: oh..this is not really used because I have defined again in the actual test which has the correct one. I will fix it anyway

test/differencer/rulesProcessor.test.ts
```diff
@@ -0,0 +1,121 @@
...
+import sinon from "sinon";
+import { ramlToolLogger } from "../../src";
+import chaiAsPromised from "chai-as-promised";
+
+/* eslint-disable @typescript-eslint/no-use-before-define*/
+
+const expect = chai.expect;
+chai.use(chaiAsPromised);
+
+const rulesPath = path.join(__dirname, "../../diffRules", "defaultRules.json");
```
wjhsf, line 22: This root `rulesPath` is shadowed in some tests, but not all, so I think it's harder to tell at a glance what `rulesPath` is in individual tests. I think it would reduce confusion if either this variable were renamed (`defaultRulesPath` or `validRulesPath` or something) or the single-use variables were renamed.

unandyala: renamed to validRulesPath

PR #54, opened 2020-06-05
Avoid using intersection types and fix wrong types.

wjhsf:
1. There are existing types that serve the same purpose as the intersection types removed.
2. The param type and return type for `resolveApiModel` were unnecessarily over-specific.
3. Cleaned up linter warnings and type annotations.
4. *Changed return type of `parseRamlFile` and `resolveApiModel` to `Document` instead of `BaseUnit`* because it's more accurate and reduces need for type assertions downstream.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -101,7 +101,7 @@ export function getAllDataTypes(
   api: model.document.BaseUnitWithDeclaresModel
 ): model.domain.CustomDomainProperty[] {
   let ret: model.domain.CustomDomainProperty[] = [];
-  const dataTypes: Set<string> = new Set();
+  const dataTypes = new Set<string>();
   const temp: model.domain.CustomDomainProperty[] = getDataTypesFromDeclare(
     api.declares,
     dataTypes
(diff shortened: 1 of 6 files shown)


wjhsf:
New changes:
1. Cleaned up linter warnings and type annotations.
2. *Changed return type of `parseRamlFile` and `resolveApiModel` to `Document` instead of `BaseUnit`* because it's more accurate and reduces need for type assertions downstream.

src/parser.ts
```diff
@@ -124,7 +125,7 @@ export function getAllDataTypes(
 export function resolveApiModel(
   apiModel: model.document.BaseUnit,
```
unandyala, line 126: why is apiModel not a Document?

src/parser.ts
```diff
@@ -161,6 +165,6 @@ export function getNormalizedName(name: string): string {
 export function getApiName(
   apiModel: model.document.BaseUnitWithEncodesModel
```
unandyala, line 166: same.. why is apiModel not a model.document.Document?

src/parser.ts
wjhsf: Because it doesn't have to be. This function never uses `encodes` or `declares`, so we don't want to use a type that requires those properties.

src/parser.ts
wjhsf: Because it doesn't have to be. This function never uses `declares`, so we don't want to use a type that requires that property.

PR #55, opened 2020-06-08
Update amf-client-js version to 4.1.2 and export a few functions

gkhunger-sf:


src/index.ts
```diff
@@ -102,7 +102,7 @@ export { extractFiles } from "./exchange-connector/exchangeDirectoryParser";
 export { RestApi } from "./exchange-connector/exchangeTypes";
 export { ramlToolLogger } from "./logger";
 
-export { model } from "amf-client-js";
+export { model, AMF, plugins } from "amf-client-js";
```
wjhsf, line 105: These aren't names we control. Should they be exported as `amf` to avoid potential name collisions?

src/index.ts
```diff
@@ -102,7 +102,7 @@ export { extractFiles } from "./exchange-connector/exchangeDirectoryParser";
 export { RestApi } from "./exchange-connector/exchangeTypes";
 export { ramlToolLogger } from "./logger";
 
-export { model, AMF, plugins } from "amf-client-js";
+export { default as amf } from "amf-client-js";
```
bredmond-sf, line 105: This breaks the SDK. I do agree overall with Will, but I'd like
```suggestion
export amf, { model } from "amf-client-js";
```
It makes sense to me export model at the top level since it is also the model we're using.

gkhunger-sf: I updated all the references in the SDK so it shouldn't break.

gkhunger-sf: I'll push the changes in commerce-sdk once these are approved.

gkhunger-sf: I couldn't find a way to do that successfully in Typescript. @bredmond-sf @wjhsf I think we will have to live with this.

wjhsf:
```suggestion
export * as amf from "amf-client-js";
export { model } from "amf-client-js";
```
I think this should do it. I don't think there's a way to consolidate that into one statement.

gkhunger-sf: Thanks Will for the suggestion.

gkhunger-sf: Please take a look and hit approve if everything looks alright.

PR #56, opened 2020-06-09
Additional diff rules

unandyala:
**Added new rules** 

> 1. Method removed
> 2. Required parameter added
> 3. Any parameter removed
> 4. Version change

- Moved the tests for the rules to a separate file (defaultRules.test.ts)
- Added additional tests to increase the coverage
- Moved the rules engine custom operators to its own ts file

### DIFF
--- a/src/diffTool/engineOperators.ts
+++ b/src/diffTool/engineOperators.ts
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Operator } from "json-rules-engine";
+
+const customOperators: Operator[] = [];
+
+customOperators.push(
+  new Operator("hasProperty", (factValue: object, jsonValue: string) =>
+    factValue.hasOwnProperty(jsonValue)
+  )
+);
+customOperators.push(
+  new Operator(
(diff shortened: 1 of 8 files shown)


src/diffTool/engineOperators.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Operator } from "json-rules-engine";
+
+const customOperators: Operator[] = [];
```
monkey-jeff, line 9: this file is "engineOperators" but only exports a "customOperators" as a default.

I remove the default and export customOperators here as a named export.  That way we support multiple types of operators.  OR change the filename/const name to match each other and maintain a single export. Otherwise, I see this can be confusing.

unandyala: I will rename it to customOperators.ts

unandyala:
> Also looking I'm confused by a couple of things.
> 
> ```
> function runEngine(engine: Engine, diff: NodeDiff): Promise<EngineResult> {
> ```
> 
> What is EngineResult and why do we throw it away?

EngineResult has same data (event and almanac) that gets passed to success handler/callback in a slightly different structure. Ideally we can use either callback or engine result to process the result

I think I could not get RuleResult which has rule name from the engine result. So I ended up using the callback. I will add a comment there and return the diffs as discussed

PR #59, opened 2020-06-10
Minor fixes in diff tool

unandyala:


monkey-jeff:
Once you resolve these conflicts I think we can move forward.

PR #61, opened 2020-06-10
reorg RAML Toolkit w 7656001

bredmond-sf:
First set of changes 
1.  enable 500 line per file limit in eslint
2. remove app.js
3. move profiles to resources/lint/profiles
4. move common files to common folder
5. add resources folder to files package.json
6. rename src/validator.ts to src/lint/lint.ts
7. move linter code from src/index.ts to src/lint/index.ts (linting should remain the default command)
8. rename src/exchange-connector to src/download


src/index.ts
```diff
@@ -4,103 +4,31 @@
...
-import fs from "fs";
-import { Command, flags } from "@oclif/command";
-import { validateFile, printResults } from "./validator";
 
-export default class RamlToolkitCommand extends Command {
-  async run(): Promise<void> {
-    const { argv, flags } = this.parse(RamlToolkitCommand);
+// This preserves the linter as the default oclif command
+import lint from "./lint";
+export default lint;
```
wjhsf, line 10:
```suggestion
export { lint as default } from "./lint";
```

bredmond-sf: Yea, it doesn't work because we're importing the default, not a named import (and oclif expects things a certain way so it needs to be default)

PR #70, opened 2020-06-17
added new function "findApiChanges" to find diffs and apply rules

unandyala:


src/diffTool/diffProcessor.ts
```diff
@@ -14,18 +14,37 @@ import { applyRules } from "./rulesProcessor";
...
+ */
+export async function findApiChanges(
+  leftRaml: string,
+  rightRaml: string,
+  rulesPath?: string
+): Promise<NodeDiff[]> {
+  const diffs = await diffRaml(leftRaml, rightRaml);
+  if (rulesPath == null) {
+    ramlToolLogger.info("Applying default rules on the differences");
+    rulesPath = path.join(diffRulesPath, "defaultRules.json");
```
wjhsf, line 33: `diffRulesPath` is a constant, and is only used here. Make the filename part of that so you don't have to recompute every time.

src/diffTool/diffProcessor.ts
```diff
@@ -12,20 +12,42 @@ import { ramlToolLogger } from "../common/logger";
...
+ * @param leftRaml - Base RAML file to compare
+ * @param rightRaml - Other RAML file to compare with the left RAML to find differences
+ * @param rulesPath - Optional rules file to apply on the differences
+ *
+ * @returns Changes to the APIs
+ */
+export async function findApiChanges(
+  leftRaml: string,
+  rightRaml: string,
+  rulesPath?: string
```
monkey-jeff, line 31: Why not `rulesPath = defaultRulesPath` or set it to the actual rule here?  Then you don't need the null check.

src/diffTool/diffProcessor.ts
```diff
@@ -12,20 +12,42 @@ import { ramlToolLogger } from "../common/logger";
...
+
+/**
+ * Generate differences between two RAML files and apply rules
+ * @param leftRaml - Base RAML file to compare
+ * @param rightRaml - Other RAML file to compare with the left RAML to find differences
+ * @param rulesPath - Optional rules file to apply on the differences
+ *
+ * @returns Changes to the APIs
+ */
+export async function findApiChanges(
```
monkey-jeff, line 28: Should this be `applyChangeRules` instead?

unandyala: There is already function apply rules that applies the rules on node diffs. 
`applyRules(diffs: NodeDiff[],rulesPath: string)`

I thought we agreed on "findApiChanges". Once we have **ApiChanges** object, its return value will also be changed to **ApiChanges**

monkey-jeff: you're right.  I retract my comment

PR #71, opened 2020-06-17
Updated difftool to use fatRamlResourceLoader

dbarbato-sf:
Updated the diffProcessor to use the internal fat raml parser.

### DIFF
--- a/src/diffTool/diffProcessor.ts
+++ b/src/diffTool/diffProcessor.ts
@@ -11,6 +11,7 @@ import { findJsonDiffs, NodeDiff } from "./jsonDiff";
 import { ramlToolLogger } from "../common/logger";
 import * as path from "path";
 import { applyRules } from "./rulesProcessor";
+import { parseRamlFile } from "../common/parser";
 
 export const diffRulesPath = path.join(__dirname, "../../resources/diff/rules");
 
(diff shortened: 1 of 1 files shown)


src/diffTool/diffProcessor.ts
```diff
@@ -49,17 +50,9 @@ export async function diffRaml(
...
+  const parsedRaml = await parseRamlFile(ramlFilePath);
 
-  //TODO: use common parser function that is used by validator and sdk
-  let model = await new amf.Raml10Parser().parseFileAsync(
-    `file://${ramlFilePath}`
-  );
-  model = new amf.Raml10Resolver().resolve(model);
+  const model = new amf.Raml10Resolver().resolve(parsedRaml);
 
   const renderOptions = new amf.render.RenderOptions().withoutSourceMaps
```
kalagarraj, line 57: are you working on tests?

monkey-jeff, line 57: Testing what? Already have test coverage on this method don't we?

kalagarraj: how abt a test that
expects parseRamlFile is called with ramlFilePath..

monkey-jeff: We actually talked about that...the problem we both see with that is that you are kind of testing the backend implementation in a way when you do that.  Which leads to a flaky test.  Basically, if we change that it would break the test but really no break anything except the test.  You want tests that when the tests break the signify some broken functionality in the code. I'm not sure doing that would do. 

 (Mind you hey said he might still write a test that does this on a branch to test out how to use sinon and such.)

src/diffTool/diffProcessor.ts
```diff
@@ -49,17 +50,9 @@ export async function diffRaml(
...
-  amf.plugins.features.AMFValidation.register();
-  await amf.Core.init();
+  const parsedRaml = await parseRamlFile(ramlFilePath);
 
-  //TODO: use common parser function that is used by validator and sdk
-  let model = await new amf.Raml10Parser().parseFileAsync(
-    `file://${ramlFilePath}`
-  );
-  model = new amf.Raml10Resolver().resolve(model);
+  const model = new amf.Raml10Resolver().resolve(parsedRaml);
```
bredmond-sf, line 55: common/parser also has a resolveApiModel function you might want to use here

src/diffTool/diffProcessor.ts
kalagarraj: I am looking this like, 

- use X instead of Y (For some reason)
- Ensure X is used as expected

monkey-jeff: But doesn't that create a test failure when there is no functionality failure?  Which really just becomes a roadblock to changing implementations out?

dbarbato-sf: Spoke with @kalagarraj offline, he’s happy to resolve this conversation.

src/diffTool/diffProcessor.ts
dbarbato-sf: I’ll take a look at this before merging.

dbarbato-sf: Given the todo around the pipelines in common/parser, I'm going to keep it like this for now. Once that todo gets resolved, this should be readdressed.

PR #72, opened 2020-06-19
Fix invalid types

wjhsf:
I noticed that this file used `JSON` as a type in a few places. That is not a valid type, it refers to the global JSON stringify/parse object. So I fixed that and a few other inaccurate type annotations. I also cleaned up a few of the functions.

I made one important change. `getSpecificApi` previously returned a Promise or null. I changed it always return a Promise (that could resolve to null), because having a mixed return type can add unnecessary complexity to the code that calls it. This change is technically a breaking change, because the function is a top-level export of the package. However, my next PR will be migrating code from commerce-sdk to here, so it will no longer need to be exported.

### DIFF
--- a/src/download/exchangeDownloader.ts
+++ b/src/download/exchangeDownloader.ts
@@ -7,11 +7,17 @@
 // import "cross-fetch/polyfill";
 
 import { writeFileSync, ensureDirSync } from "fs-extra";
-
+import { clone } from "lodash";
 import fetch, { Response } from "node-fetch";
 import path from "path";
 
-import { RestApi, FileInfo, Categories } from "./exchangeTypes";
+import {
+  RawRestApi,
+  RestApi,
+  FileInfo,
+  RawCategories,
+  Categories
+} from "./exchangeTypes";
 import { ramlToolLogger } from "../common/logger";
 
 const DEFAULT_DOWNLOAD_FOLDER = "download";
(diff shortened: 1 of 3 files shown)


src/download/exchangeDownloader.ts
```diff
@@ -40,57 +45,49 @@ export async function downloadRestApi(
...
+function getFileByClassifier(files: FileInfo[], classifier: string): FileInfo {
+  const found = files.find(file => file.classifier === classifier);
+  // There are extra properties we don't want (downloadURL, isGenerated), so we
+  // create a new object that excludes them
+  return {
+    classifier: found.classifier,
+    packaging: found.packaging,
+    externalLink: found.externalLink,
+    createdDate: found.createdDate,
+    md5: found.md5,
```
monkey-jeff, line 74: Do we need to keep these elements in the original object at this point? would it make more sense to do this instead basically?

```typescript
delete found.downloadURL;
delete found.isGenerated;
return found;
```

?

wjhsf: Is there a reason why they are excluded in the first place?

monkey-jeff: Download Url is excluded because it expires after 24 hours so it shouldn't be there. Also as it is a presigned s3 url it contains creds to download it from s3 for 24 hours and I didn'nt want to save that in git as bots will pick it up. I forget why isGenerated was ecluded

PR #73, opened 2020-06-22
Br/diff tool cli.w 7435293

bredmond-sf:
AC:
Diff tool can be called from CLI directly.
Offer options to specify the two RAML files
Two is the number of RAML files that must be provided
No conflicts with existing linting commands
Refactor linter to be a sub-command and make new capability also a sub-command
Default to linter behavior

To try, clone and `npm run compile && bin/run diff --help`
Run unit tests with `npx mocha src/diff/index.test.ts`


bin/run
```diff
@@ -6,17 +6,14 @@
...
-  .default.run()
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  .catch(require("@oclif/errors/handle"));
+require("@oclif/command")
+  .run()
+  .then(require("@oclif/command/flush"))
+  .catch(e => {
+    if (e.message.startsWith("command")) {
+      return require("../lib/commands/lint")
+        .default.run();
```
monkey-jeff, line 15: [BLOCKER] Not providing the ts-node register means that you must recompile on all changes before testing.  It also means I think that debugging the typescript directly will be more difficult since it will be going through the compiled javascript.

bin/run
```diff
@@ -6,17 +6,14 @@
...
-// require(`../${dev ? "src" : "lib"}`)
-require(`../${dev ? "src" : "lib"}`)
-  .default.run()
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  .catch(require("@oclif/errors/handle"));
+require("@oclif/command")
+  .run()
+  .then(require("@oclif/command/flush"))
+  .catch(e => {
+    if (e.message.startsWith("command")) {
```
wjhsf, line 13: Maybe add a comment explaining the reasoning here?

src/diff/index.ts
```diff
@@ -0,0 +1,99 @@
...
+        args.apiSpecNewPath,
+        flags.ruleset
+      );
+      console.log(results);
+    } catch (err) {
+      this.error(err.message, { exit: 2 });
+    }
+
+    // TODO: Move to logic to the library
+    const hasBreakingChanges = (diff: NodeDiff[]): boolean =>
```
monkey-jeff, line 90: Hmm, is tehre a reason you made this an inline method with a todo statement to move the logic instead of just moving the logic to the library?

src/diff/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      diff.some(n => n.rule?.params?.category === "Breaking");
+
+    if (hasBreakingChanges(results)) {
+      this.error("Breaking changes found.", { exit: 1 });
+    }
+
+    this.exit();
```
monkey-jeff, line 97: Is this necessary?

src/diff/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    },
+    {
+      name: "apiSpecNewPath",
+      required: true,
+      description:
+        "The new version of the API spec for comparison against the base version"
+    }
+  ];
+
+  async run(): Promise<void> {
```
monkey-jeff, line 56: Not a huge fan of the flow of this.  Kind of flows like a switch statement with ifs, where they are mutually exclusive by simple virtue of exits at the end of the if.  Would it make more sense to convert to a switch statment here and hole exit codes until the end?

bin/run
bredmond-sf: That's actually a shortcoming in oclif. The default wrapper for a multi-command cli is ``` /tmp/mynewcli      cat bin/run        
#!/usr/bin/env node

require('@oclif/command').run()
.then(require('@oclif/command/flush'))
.catch(require('@oclif/errors/handle'))
```
They only have that dev-mode support for single command CLI's. Let me think if there's a simple way we might try to have something similar, but it's not something oclif provides support for.

File: src/diff/index.ts [hunk 6]
```diff
@@ -0,0 +1,99 @@
...
+    const { args, flags } = this.parse(DiffCommand);
+
+    let results: NodeDiff[];
+
+    // Don't apply any ruleset, exit 0 for no differences, exit 1 for any
+    // differences, exit 2 for unsuccessful
+    if (flags["diff-only"]) {
+      try {
+        results = await diffRaml(args.apiSpecBasePath, args.apiSpecNewPath);
+        console.log(results);
```
wjhsf, line 66:
```suggestion
        this.log(results);
```
Oclif's built-in logger.

src/diff/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    // Apply ruleset, exit 0 for no breaking changes, exit 1 for breaking
+    // changes, exit 2 for unsuccessful
+    try {
+      results = await findApiChanges(
+        args.apiSpecBasePath,
+        args.apiSpecNewPath,
+        flags.ruleset
+      );
+      console.log(results);
```
wjhsf, line 84:
```suggestion
      this.log(results);
```

src/diff/index.ts
bredmond-sf: Because you and @unandyala have been making changes in the library so I wanted to get this open and figure out how we can do this

src/diff/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+import cmd from "../commands/diff";
+
+const createTempFile = (content: string): FileResult => {
+  const tempFile = fileSync();
+  writeSync(tempFile.fd, content);
+  return tempFile;
+};
+
+const ramlOld = createTempFile(
```
wjhsf, line 21: 1) Why are tmp files needed?
2) Maybe consider having these in their own files and copy them to a tmp dir for these tests.

src/diff/index.ts
bredmond-sf: Strictly speaking no, but without it, oclif-test sees the exit code as undefined instead of 0

src/diff/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+/resource:
+  get:
+    displayName: getResource
+  post:
+    displayName: createResource
+`
+);
+
+const operationRemovedRule = createTempFile(
+  `[{
```
unandyala, line 49: consider using Rule class provided by json-rules-engine. 
Example: 
```
const r = new Rule(ruleProperties);
r.toJSON() //gives the json string
```

src/diff/index.ts
bredmond-sf: This will get replaced once there's a formatter, but for now, console.log nicely formats the output, unlike this.log which only accepts a string.

src/diff/index.test.ts
bredmond-sf: These are tests of the CLI and not the diff functions themselves (those have their own tests) so I prefer having these inline since they're meant to be as simple as possible.

src/diff/index.test.ts
bredmond-sf: done

src/diff/index.ts
bredmond-sf: I don't love it either, but I'm staring at it and struggling to find something I like better. I can't see where a switch would come in...there's no place here where there's three branches based on a single value. There's one behavior for diff-only and a different one without it. I can break it up, but it's a single screen of code so it just feels like more jumping around to follow it. Open to suggestions, I'm just not seeing something that reads easier to me.

src/diff/index.ts
```diff
@@ -0,0 +1,99 @@
...
+  2 - evaluation could not be completed
+
+The ruleset flag is used to evaluate a custom ruleset in place of the default rules. The diff-only flag disables evaluation against any ruleset.
+`;
+
+  static flags = {
+    // Add --version flag to show CLI version
+    version: flags.version({ char: "v" }),
+    // Add --help flag to show CLI version
+    help: flags.help({ char: "h" }),
```
wjhsf, line 28:
```suggestion
```
These are automatically supported.

bin/run
bredmond-sf: I put together a quick wrapper as bin/run-dev that gives similar function for multicommand

src/diff/index.ts
bredmond-sf: Version doesn't work without this here and help only work in long form without it.

src/diff/index.ts
monkey-jeff: Still don't love the flow here and think we should be thinking about what we want to do for these kinds of main flow logic so it flows a little more smoothly.

PR #75, opened 2020-06-25
Add lint:dev command

wjhsf:
Same as https://github.com/SalesforceCommerceCloud/commerce-sdk/pull/216

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
       "ecmaVersion": 2018,
       "sourceType": "module"
     },
+    "reportUnusedDisableDirectives": true,
     "rules": {
       "@typescript-eslint/interface-name-prefix": [
         "error",
@@ -121,7 +122,8 @@
     "postpack": "rm -f oclif.manifest.json",
     "prepack": "npm run clean && npm run compile && oclif-dev manifest && oclif-dev readme",
     "pretest": "npm run lint && npm run compile && npm run depcheck",
-    "lint": "eslint \"**/*.ts\" --quiet --fix",
+    "lint": "npm run lint:dev -- --quiet",
+    "lint:dev": "eslint . --ext .ts --fix",
     "test:unit": "nyc mocha \"src/**/*.test.ts\" -- --reporter=xunit --reporter-options output=reports/generator.xml",
     "test:debug": "mocha \"src/**/*.test.ts\"",
     "test:resources": "mocha \"resources/**/*.test.ts\"",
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -121,7 +122,8 @@
     "postpack": "rm -f oclif.manifest.json",
     "prepack": "npm run clean && npm run compile && oclif-dev manifest && oclif-dev readme",
     "pretest": "npm run lint && npm run compile && npm run depcheck",
-    "lint": "eslint \"**/*.ts\" --quiet --fix",
+    "lint": "npm run lint:dev -- --quiet",
```
monkey-jeff, line 125: Will this now run against compiled javascript code? Why drop the ts filter here?

wjhsf: As far as I can tell, `eslint "**/*.ts"` and `eslint . --ext .ts` are equivalent. Went with the latter for consistency with commerce-sdk and because I think avoiding the escaped quotes makes it easier to read in the package.json.

monkey-jeff: Yes...but you didn't add --ext ts to this command.  Only the dev version.

monkey-jeff: nevermind I see what you are doing now.

resources/lint/profiles/mercury.base-uri.test.ts
```diff
@@ -23,7 +22,6 @@ describe("base uri", () => {
       "http://a.ml/vocabularies/data#base-uri-matches-pattern";
     const TEMPLATE_RULE =
       "http://a.ml/vocabularies/data#camelcase-template-parameters";
-    const TITLE_RULE = "http://a.ml/vocabularies/data#title-matches-api-name";
```
dbarbato-sf, line 26: If this is being removed as it's unused, should it still be present and we're missing functionality?
If so, please create a new story to address it.

src/diff/jsonDiff.ts
```diff
@@ -335,7 +334,6 @@ export function addNodeDiff(diff: object[]): NodeDiff {
  */
 export function addNodePropertyDiffs(
   key: string,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dbarbato-sf, line 338: Question, why were these calls not failing linting previously?

wjhsf: Updated the file to avoid using a lot of `any`s. I replaced them with `unknown`s, which is not ideal, but better types will hopefully be added in the coming conversion to OO style.

resources/lint/profiles/mercury.base-uri.test.ts
wjhsf: @gkhunger-sf looks like you added this in #26.

gkhunger-sf: We never added this rule. I opened a PR and found that this rule is going away in the next api spec. I don't know how it ended up here. Thanks for removing it, Will.

PR #76, opened 2020-07-02
Create Api data models Br/add api models.w 7768368

bredmond-sf:
AC:
API classes described in the data model section of this doc are added to and exported in raml-toolkit
Add Instructions on using these data models in the README

Test Plan: Unit Tests `npx mocha src/common/api*`


src/generate/apiCollection.ts
```diff
@@ -0,0 +1,31 @@
...
+   *
+   * @param description - a simple object literal with group names as
+   * properties and a list of file paths under each group
+   */
+  static async read(description: {
+    [key: string]: string[];
+  }): Promise<ApiCollection> {
+    const apiCollection = new ApiCollection();
+    for (const group in description) {
+      apiCollection[group] = await ApiGroup.read(description[group]);
```
wjhsf, line 26: Do we have to set groups directly on the instance? Setting a property might be safer (avoids naming collisions with methods). Something like `apiCollection.collection = new Map()` or `apiCollection.collection = Object.create(null)`

wjhsf: Additionally, if this change is made, then this class could also get `name`/`setName`, which I think would be useful.

src/generate/apiGroup.ts
```diff
@@ -0,0 +1,61 @@
...
+   * constructor because it is async. Files are processed in parallel.
+   *
+   * @param apiSpecFilePaths - a list of paths to API spec files like RAML
+   */
+  static async read(apiSpecFilePaths: string[]): Promise<ApiGroup> {
+    const apiPromises: Promise<Api>[] = [];
+    for (const apiSpecFilePath of apiSpecFilePaths) {
+      apiPromises.push(Api.read(apiSpecFilePath));
+    }
+    await Promise.all(apiPromises);
```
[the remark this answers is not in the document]
bredmond-sf: *That's* the thing I was looking for. The contrived code here is to convert from type of Promise to specific value. Fixed.

src/generate/api.ts
```diff
@@ -0,0 +1,63 @@
...
+  };
+  path: string;
+
+  /**
+   * Create a new Api object from a file. This is static and not a constructor
+   * because it is async.
+   *
+   * @param apiSpecFilePath - the path to an API spec file like RAML to be parsed to AMF
+   */
+  static async read(apiSpecFilePath: string): Promise<Api> {
```
monkey-jeff, line 39: This seems like a factory pattern and the name should probably reflect that

src/generate/api.ts
```diff
@@ -0,0 +1,63 @@
...
+    const api = new Api();
+    api.path = apiSpecFilePath;
+    api.model = resolveApiModel(
+      await parseRamlFile(apiSpecFilePath),
+      "editing"
+    );
+    api.setName((api.model.encodes as model.domain.WebApi).name.value());
+    api.dataTypes = getAllDataTypes(api.model);
```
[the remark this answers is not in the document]
monkey-jeff: I agree some of these should be moved into the constructor but use your judgment here brian.

src/generate/api.ts
```diff
@@ -0,0 +1,63 @@
...
+  parseRamlFile,
+  getAllDataTypes
+} from "../common/parser";
+
+/**
+ * An API represented as an AMF model. Includes the extracted data types
+ * defined in the spec. Common transformations of the name are cached for
+ * reference in templates and for use in assembling file paths.
+ */
+export class Api {
```
monkey-jeff, line 21: MIght want a field pointing back at the group.  That way an api knows what group it is a part of.

src/generate/apiCollection.ts
```diff
@@ -0,0 +1,31 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ApiGroup } from "./apiGroup";
+
+/**
+ * A collection of multiple groups of APIs stored by key.
+ */
+export class ApiCollection {
```
monkey-jeff, line 12: One thing to just keep in mind...we group the APIs in the sdk we've built because of the sheer number....but we should definitely support a flat structure as well at some point.

src/generate/apiGroup.ts
```diff
@@ -0,0 +1,51 @@
...
+   * Create a group from a list of API spec files. This is static and not a
+   * constructor because it is async. Files are processed in parallel.
+   *
+   * @param apiSpecFilePaths - a list of paths to API spec files like RAML
+   */
+  static async read(apiSpecFilePaths: string[]): Promise<ApiGroup> {
+    return new ApiGroup("", await Promise.all(apiSpecFilePaths.map(Api.read)));
+  }
+
+  setName(name = ""): ApiGroup {
```
wjhsf, line 40:
```suggestion
  setName(name = ""): this {
```
This is a thing.

wjhsf: (Mostly just for the comment. 😁)

bredmond-sf: I like it, updating

src/generate/apiCollection.ts
bredmond-sf: Absolutely. I don't see us making anyone use this class if they don't need it; they can just use a single instance of ApiGroup.

src/index.ts
```diff
@@ -38,4 +38,8 @@ export {
   resolveApiModel
 } from "./common/parser";
 
+export { Api } from "./generate/api";
```
unandyala, line 41: How about having index.ts inside generate directory similar to diff and lint

src/generate/api.ts
bredmond-sf: I like this on the surface, but then I started to think about the code to manage keeping it in sync and I don't love that part (we'd have to wrap up the array in ApiGroup). How much value do you think it is?

src/generate/api.ts
bredmond-sf: It's factory-ish, but it only ever returns one type of object, so it's not actually. I'm open to different names, but I think it might be confusing to call it factory since it's not really.

wjhsf: load / loadFromFile(s)?

src/generate/api.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+before(() => {
+  chai.use(chaiAsPromised);
+});
+
+describe("Test Api class read", () => {
+  it("creates an instance from a valid raml file", async () => {
+    const api = await Api.read(validRamlFile);
+    expect(api.dataTypes).to.not.be.empty;
+    expect(api.model).to.not.be.empty;
```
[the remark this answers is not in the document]
bredmond-sf: I agree with the statement, but in this case these assertions are covered by TypeScript.

src/generate/api.ts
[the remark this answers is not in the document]
bredmond-sf: doing it

src/generate/api.ts
monkey-jeff: Well let me rephrase my concern and you let me know how to solve it.

First...Im not too concerned about drift here since we aren't rewriting data much...once it is read it is fairly immutable for that generation run.

Secondly, if you look what I was doing around inline documentation and example code we need to have some bread crumb of where it came from.  This is the primary aspect I am trying to solve.

src/generate/api.ts
```diff
@@ -5,30 +5,56 @@
...
-  model: model.document.Document;
-  name: {
-    original: string;
-    kebabCase: string;
-    lowerCamelCase: string;
-    snakeCase: string;
-    upperCamelCase: string;
-  };
-  path: string;
+export default class Api {
```
unandyala, line 22: I was debating whether to use default on the class. Default exports have maintainability  issues

https://basarat.gitbook.io/typescript/main-1/defaultisbad

wjhsf: TypeScript convention (according to that Udemy course) is to always use named exports.

src/generate/api.ts
monkey-jeff: I like "create" or "make" somewhere in the name, or something that signifies the creation of something is taking place.  Like `createApiFromFile` or just something like `makeApi` or `modelApi` or 'createApiModel`

wjhsf: I think that using the class name in the function name is redundant (and uncommon; see `Handlebars.create()`). Using a creation verb is a good idea.

src/generate/apiCollection.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect(collection["Group One"].apis[0].model).to.not.be.empty;
+    expect(collection["Group One"].apis[0].name.original).to.equal("Shop API");
+    expect(collection["Group One"].apis[0].path).to.be.equal(validRamlFile);
+  });
+
+  it("creates an instance from two valid raml files in one group", async () => {
+    const collection = await ApiCollection.read({
+      "Group One": [validRamlFile, validRamlFile]
+    });
+    expect(collection["Group One"].apis.length).to.eq(2);
```
wjhsf, line 36:
```suggestion
    expect(collection["Group One"].apis).to.be.an("array").with.lengthOf(2);
```
Chai provides array-specific assertions.

src/generate/name.test.ts
```diff
@@ -0,0 +1,111 @@
...
+describe("Test Name class", () => {
+  it("sets all fields to empty string from the default constructor", () => {
+    const name = new Name();
+    expect(name.original).to.equal("");
+    expect(name.kebabCase).to.equal("");
+    expect(name.lowerCamelCase).to.equal("");
+    expect(name.snakeCase).to.equal("");
+    expect(name.upperCamelCase).to.equal("");
+    expect(name.toString()).to.equal("");
+    expect(name + "T").to.equal("T");
```
wjhsf, line 20:
```suggestion
```
String concatenation always calls `toString()`, so these tests are redundant.

src/generate/name.ts
```diff
@@ -0,0 +1,54 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import _ from "lodash";
+
+/**
+ * Stores a name with common transformations cached for use in templates and file paths.
+ */
+export default class Name {
```
wjhsf, line 12: Do we expect people to be modifying names all that often? I think it's probably fine to just use normal properties.

src/generate/api.ts
```diff
@@ -0,0 +1,71 @@
...
+  set model(model: model.document.Document) {
+    this._model = model;
+    this._dataTypes = getAllDataTypes(model);
+    this._name = new Name((model.encodes as model.domain.WebApi)?.name.value());
+  }
+
+  get name(): Name {
+    return this._name;
+  }
+
```
wjhsf, line 54:
```suggestion

set name(name: string | Name): void {
  this._name = new Name(name?.toString() || "");
}

```
Allows users to do `api.name = "..."` instead of `api.name.original = "..."`.

src/generate/api.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * An API represented as an AMF model. Includes the extracted data types
+ * defined in the spec. Common transformations of the name are cached for
+ * reference in templates and for use in assembling file paths.
+ */
+export default class Api {
+  private _dataTypes: model.domain.CustomDomainProperty[] = [];
+  private _model = new model.document.Document();
+  private _name = new Name();
+  private _path = "";
```
wjhsf, line 26: Why are these private variables with getters?

bredmond-sf: All of the fields here are somewhere in the model. I can't come up with a good reason to not keep them in sync.

src/generate/name.ts
bredmond-sf: This makes it so the transformations stay in sync with the original field

src/generate/name.test.ts
bredmond-sf: true, removed

src/generate/apiCollection.test.ts
bredmond-sf: updated

src/generate/api.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+before(() => {
+  chai.use(chaiAsPromised);
+});
+
+describe("Test Api class read", () => {
+  it("constructs an instance with no args", () => {
+    const api = new Api();
+    expect(api.dataTypes).to.be.empty;
+    expect(api.model).to.eql(new model.document.Document());
```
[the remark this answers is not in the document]
bredmond-sf: done

docs/generate/generate.md
```diff
@@ -0,0 +1,39 @@
...
+
+## API Classes
+
+The API classes in this package provide a simple interface to read one or an entire collection of RAML files. The instances are then ready to use in [Handlebars](https://www.npmjs.com/package/handlebars) templates (Handlebars is an extension to the popular Mustache templating language).
+
+For a single RAML file:
+```javascript
+import { Api } from "@commerce-apps/raml-toolkit";
+
+Api.read("path/to/my-spec.raml").then(console.log);
```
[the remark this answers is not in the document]
bredmond-sf: done

src/generate/api.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor(amfModel = new model.document.Document(), path = "") {
```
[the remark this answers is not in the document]
bredmond-sf: nope, removed

src/generate/api.ts
bredmond-sf: Api now has a string field for 'group' that is set when init'ing group

src/index.ts
bredmond-sf: done

src/generate/apiGroup.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A group of API objects. Common transformations of the group name are cached
+ * for reference in templates and file paths.
+ */
+export class ApiGroup {
+  apis: Api[];
+  name: Name;
+
+  constructor(name = "", apis?: Api[]) {
+    this.name = new Name(name);
```
wjhsf, line 21: A user could naively do `group.name = "My Group"`, which could break things. 🤷

src/generate/apiGroup.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.apis = apis;
```
wjhsf, line 22:
```suggestion
    this.apis = apis || [];
```
The constructor parameter is optional, but the class definition requires it to be set. (Alternative to this is to update the parameter or class definition.)

src/generate/apiGroup.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  it("sets the name for lowercase", () => {
+    const group = new ApiGroup("This is my test name.");
+    expect(group.name.original).to.equal("This is my test name.");
+    expect(group.name.kebabCase).to.equal("this-is-my-test-name");
+    expect(group.name.lowerCamelCase).to.equal("thisIsMyTestName");
+    expect(group.name.snakeCase).to.equal("this_is_my_test_name");
+    expect(group.name.upperCamelCase).to.equal("ThisIsMyTestName");
+  });
```
wjhsf, line 74:
```suggestion
```
These tests are no longer necessary, as the `setName` method was removed.

src/generate/apiCollection.ts
```diff
@@ -0,0 +1,28 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ApiGroup } from "./apiGroup";
+
+/**
+ * A collection of multiple groups of APIs stored by key.
+ */
+export class ApiCollection extends Map {
```
wjhsf, line 12: You said that you were going to change back to not extending Map. Did you change your mind?

src/generate/apiGroup.ts
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import _ from "lodash";
+
```
wjhsf, line 8:
```suggestion
```

src/generate/apiGroup.ts
bredmond-sf: But the field type is Name

src/generate/apiGroup.ts
bredmond-sf: updated parameter

src/generate/apiGroup.test.ts
bredmond-sf: still needed but corrected the description

src/generate/apiCollection.ts
bredmond-sf: Ha, so far, yes. They're still iterable in Handlebars which I think is enough.

PR #77, opened 2020-07-03
[jb/test-circle] Test that Circle

jbegleiter-sf:
- Don't be square


wjhsf:
What a roller coaster!

PR #78, opened 2020-07-06
Add download command

wjhsf:
Added new `raml-toolkit download` command. Accepts the following flags:

| Flag | Char | Env Var | Default | Description |
| --- | --- | --- | --- | --- |
| `--search` | `-s` | `EXCHANGE_SEARCH` | `'category:"CC Visibility" = "External"'` | Search query to filter results from Exchange |
| `--deployment` | `-D` | `EXCHANGE_DEPLOYMENT` | `"production"` | Deployment status to filter results from Exchange |
| `--deployment-regex-flags` |  |  | `"i"` | RegExp flags to specify for advanced deployment matching |
| `--dest` | `-d` | `EXCHANGE_DOWNLOAD_DEST` | `"apis"` | Directory to download APIs into |
| `--family` | `-f` | `API_FAMILY` | `"CC API Family"` | The category to use to group APIs together |
| `--config-file` | `-c` | `API_CONFIG_FILE` | `"api-config.json"` | Name of the target file to save the API config |

The command also requires the environment variables `ANYPOINT_USERNAME` and `ANYPOINT_PASSWORD` to be set.

The default values all mimic the current behavior of `updateApis` in `commerce-sdk`.

To run locally, check out the branch and run `./bin/run-dev download`.

I'd particularly appreciate feedback on names/descriptions/logging (none at present, should probably be added).

Tests are failing because of code coverage, and I'm not sure why it thinks I haven't tested the command. :|


src/download/command.ts
```diff
@@ -0,0 +1,87 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import fs from "fs-extra";
+import path from "path";
+import { Command, flags } from "@oclif/command";
+import { search, downloadRestApis } from "./exchangeDownloader";
+import { extractFiles } from "./exchangeDirectoryParser";
+import { groupByCategory, removeRamlLinks } from "./exchangeTools";
+
+export default class DownloadCommand extends Command {
```
unandyala, line 14: consider avoiding default export

src/download/command.ts
```diff
@@ -0,0 +1,87 @@
...
+    }),
+    "deployment-regex-flags": flags.string({
+      description: "RegExp flags to specify for advanced deployment matching",
+      default: "i",
+      dependsOn: ["deployment"]
+    }),
+    dest: flags.string({
+      char: "d",
+      description: "Directory to download APIs into",
+      env: "EXCHANGE_DOWNLOAD_DEST",
```
unandyala, line 41: Do we need "EXCHANGE" prefix for every option? As the command itself is "downloading from exchange", we can probably get rid of the prefix

src/download/command.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+      );
+    }
+    const { flags } = this.parse(DownloadCommand);
+    if (flags["config-file"] !== path.basename(flags["config-file"])) {
+      this.error("Config file name cannot be a path.", {
+        exit: 2
+      });
+    }
+    const apis = await search(
```
unandyala, line 72: do we need try-catch around these call to throw a meaningful error to the user?

README.md
```diff
@@ -31,6 +31,8 @@ OPTIONS
   -w, --warnings                   Show all the warnings
 ```
 
+> **Note:** Some commands require environment variables to be set. This can be done using a [.env file](https://www.npmjs.com/package/dotenv#rules) in your working directory (the directory from which you run `raml-toolkit`).
+
 ### Jenkins
```
unandyala, line 36: CLI usage doc?

src/download/command.ts
```diff
@@ -0,0 +1,87 @@
...
+      flags.search,
+      new RegExp(flags.deployment, flags["deployment-regex-flags"])
+    );
+    await downloadRestApis(apis, flags.dest);
```
bredmond-sf, line 76: Around the question of logging, I personally like to see a log of each file written. Having a --quiet flag probably also makes sense

wjhsf: 🤔 A flag to control log level would probably be a useful thing for all commands. I'll stick to just adding logging for this PR, but add a story to add a common flag.

.depcheckrc.yml
```diff
@@ -1,12 +1,2 @@
-ignores: [
-  "@oclif/dev-cli",
-  "depcheck",
-  "eslint*",
-  "mocha",
-  "nyc",
-  "prettier",
-  "snyk",
-  "sort-package-json",
```
bredmond-sf, line 9:
``` 
~/src/raml-toolkit-will  wjh/download-cmd.W-7710840  0m  npm run depcheck                                                                      

> @commerce-apps/raml-toolkit@0.4.1 depcheck /Users/bredmond/src/raml-toolkit-will
> depcheck

Unused devDependencies
* sort-package-json
* tslib
```
The tslib one doesn't trigger after `npm run compile`, but the sort-package-json still does.

monkey-jeff: Why does this pass on circle then?

bredmond-sf: Hmmm, maybe we upgraded depcheck. I manually deleted my node_modules, ran npm install and now it's not complaining. I'm guessing that's why none of the other ignores are needed now as well.

PR #79, opened 2020-07-06
Template class implementation

unandyala:
- Template class holds the template and renders the template with the given context/data
- Accepts custom Handlebars environment for the rendering

Note: Need to move Template class and tests to generate package which will be added by @bredmond-sf PR (https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/76)


src/common/template.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export class Template {
+  content: string;
+
+  /**
+   * Creates a Template object
+   * @param path - Path to the template file
+   * @param hb - Optional handlebars instance, default instance is used if not provided
+   */
+  constructor(public path: string, public hb = Handlebars) {
```
wjhsf, line 21:
```suggestion
  constructor(public path: string, public handlebars = Handlebars) {
```
Using the full name makes it more clear what the property actually is.

src/common/template.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import fs from "fs-extra";
+import Handlebars from "handlebars";
```
wjhsf, line 8: Currently the class uses the default/global Handlebars instance. I thought that yesterday we had decided on using a custom instance? (So that we don't have accidental collisions if someone is using the global instance elsewhere.)

src/common/template.test.ts
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { expect } from "chai";
+import { Template } from "./template";
+import { fileSync } from "tmp";
+import { writeFileSync, readFileSync } from "fs-extra";
```
wjhsf, line 10: In my PR, I've added the `chai-fs` plugin. You should consider using it, I've found it quite helpful.

src/common/template.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    );
+  });
+  it("creates instance of template with a valid template file", async () => {
+    const templateFile = fileSync({ postfix: ".hbs" });
+    const templateContent = "Test";
+    writeFileSync(templateFile.name, templateContent);
+    const template = new Template(templateFile.name);
+    //verify template and its content
+    expect(template).to.be.an.instanceof(Template);
+    expect(template.content).to.equal(templateContent);
```
wjhsf, line 40:
```suggestion
    expect(template.content).to.equal("Test");
```
Test values should be hard-coded when possible, to avoid weird/unexpected side effects.

src/common/template.test.ts
```diff
@@ -0,0 +1,82 @@
+/*
```
wjhsf, line 1: Great test descriptions in this file!

src/common/template.ts
unandyala: I felt there are too many "handlebars" which is kind of confusing

src/common/template.ts
```diff
@@ -0,0 +1,49 @@
...
+  content: string;
+
+  /**
+   * Creates a Template object
+   * @param path - Path to the template file
+   * @param hb - Optional handlebars instance, default instance is used if not provided
+   */
+  constructor(public path: string, public hb = Handlebars) {
+    try {
+      this.content = fs.readFileSync(path, "utf8");
```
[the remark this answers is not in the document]
unandyala: I think we should allow/support one encoding through out the code so that we avoid conversion issues in future

src/common/template.ts
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @param context - The data model to use for substitutions
+   * @param destination - The file path to write the rendered template to
+   */
+  render(context: { [key: string]: unknown }, destination: string): void {
+    fs.outputFileSync(
+      destination,
+      //Parts of the AMF model use prototype properties and methods, we need to make those available to Handlebars
+      this.handlebars.compile(this.content)(context, {
+        allowProtoPropertiesByDefault: true,
```
monkey-jeff, line 44: Should we expose this configuration?

src/common/template.ts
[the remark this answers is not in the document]
monkey-jeff: I am okay with this.  I don't really see us supporting anything but UTF-8 for a long time here.

src/common/template.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  /**
+   * Make substitutions and write the template to a file.
+   *
+   * @param context - The data model to use for substitutions
+   * @param destination - The file path to write the rendered template to
+   */
+  render(context: { [key: string]: unknown }, destination: string): void {
```
monkey-jeff, line 39: Should this be returning void?

src/common/template.ts
unandyala: probably, we can expose both compile and runtime options

src/common/template.ts
unandyala: yes, it renders to the given destination.

src/common/template.ts
bredmond-sf: Can you think of a scenario where someone would need to worry about malicious templates?

src/common/template.ts
monkey-jeff: Is it synchronous?

monkey-jeff: (And should it be synchronous if it is?)

src/common/template.ts
bredmond-sf: My personal preference is that this be minimal configuration and if someone needs something more advanced to let them fallback to using Handlebars directly.

unandyala: Agree, as the current implementation of Template is just a wrapper. If Template is going to have more logic/features (just saying, not sure it will) in future then probably it make sense to expose these options

src/common/template.ts
unandyala: its synchronous right now

src/common/template.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    //render template
+    const data = { name: "Test rendering" };
+    const renderedFile = fileSync();
+    template.render(data, renderedFile.name);
+
+    //verify the rendered content
+    expect(renderedFile.name)
+      .to.be.a.file()
+      .with.content(data.name.toUpperCase());
+  });
```
gkhunger-sf, line 87: Can we have a check to ensure that the passed customHb instance is actually used?

unandyala: there is helper registered with the custom hb to ensure it is used

package.json
```diff
@@ -39,7 +39,8 @@
     "tmp": "^0.2.0",
     "ts-node": "^8.9.1",
     "typescript": "^3.7.5",
-    "unzipper": "^0.10.11"
+    "unzipper": "^0.10.11",
+    "handlebars": "^4.7.6"
```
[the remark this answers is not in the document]
bredmond-sf: Haha I almost made the same comment, but then I added a command to my PR that sorts everything in package.json automatically

src/common/template.ts
monkey-jeff: Should it be synchronous?

unandyala: keeping it synchronous lets us handle any errors better (I added try catch around it)

wjhsf: The error handling is the same for sync/async. All you'd need to change here is `fs.outputFileSync` to `await fs.outputFile` and the return type to `Promise<void>`.

src/common/template.test.ts
gkhunger-sf: my bad

PR #83, opened 2020-07-09
Prevent ts files showing as profile options.

wjhsf:
`raml-toolkit lint --help` showed typescript files as options for the profile flag. Now it doesn't.

### DIFF
--- a/src/lint/index.ts
+++ b/src/lint/index.ts
@@ -14,6 +14,11 @@ export const profilePath = path.join(
   "../../resources/lint/profiles"
 );
 
+const profiles = fs
+  .readdirSync(profilePath)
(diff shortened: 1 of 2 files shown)


src/lint/index.ts
```diff
@@ -14,6 +14,11 @@ export const profilePath = path.join(
   "../../resources/lint/profiles"
 );
 
+const profiles = fs
```
monkey-jeff, line 17: This can work in the short term, but I don't love it for a long term solution.

We should have a method for registering new profiles to the linter that makes sense.  While this works I just don't see it as scalable. I think someone should investigate a better way of doing this in the future.

PR #84, opened 2020-07-09
Make ApiCollection extend Array

bredmond-sf:
Support for Map in Handlebars is very poor, so this changes ApiCollection to extend Array.

### DIFF
--- a/src/generate/api.ts
+++ b/src/generate/api.ts
@@ -12,7 +12,7 @@ import {
   getAllDataTypes
 } from "../common/parser";
 
-import Name from "./name";
+import { Name } from "./name";
 
 /**
  * An API represented as an AMF model. Includes the extracted data types
--- a/src/generate/api.ts
+++ b/src/generate/api.ts
@@ -21,17 +21,15 @@ import { Name } from "./name";
  */
 export class Api {
   dataTypes: model.domain.CustomDomainProperty[];
-  group: string;
-  model: model.document.Document;
   name: Name;
-  path: string;
 
-  constructor(model, group = "", path = "") {
+  constructor(
+    public model: model.document.Document,
+    public group = "",
+    public path = ""
+  ) {
     this.dataTypes = getAllDataTypes(model);
-    this.group = group;
-    this.model = model;
     this.name = new Name((model.encodes as model.domain.WebApi)?.name.value());
-    this.path = path;
     Object.freeze(this);
   }
 
--- a/src/generate/apiCollection.test.ts
+++ b/src/generate/apiCollection.test.ts
@@ -23,19 +23,17 @@ describe("Test ApiGroup class init", () => {
     const collection = await ApiCollection.init({
       "Group One": [validRamlFile]
     });
-    expect(collection.get("Group One").apis).to.not.be.empty;
-    expect(collection.get("Group One").apis[0].model).to.not.be.empty;
-    expect(collection.get("Group One").apis[0].name.original).to.equal(
-      "Shop API"
-    );
-    expect(collection.get("Group One").apis[0].path).to.be.equal(validRamlFile);
+    expect(collection.get("Group One")).to.not.be.empty;
+    expect(collection.get("Group One")[0].model).to.not.be.empty;
+    expect(collection.get("Group One")[0].name.original).to.equal("Shop API");
+    expect(collection.get("Group One")[0].path).to.be.equal(validRamlFile);
   });
 
   it("creates an instance from two valid raml files in one group", async () => {
     const collection = await ApiCollection.init({
       "Group One": [validRamlFile, validRamlFile]
     });
-    expect(collection.get("Group One").apis)
+    expect(collection.get("Group One"))
       .to.be.an("array")
       .with.lengthOf(2);
   });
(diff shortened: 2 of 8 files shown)


src/generate/apiCollection.ts
```diff
@@ -19,10 +19,14 @@ export class ApiCollection extends Map {
...
-    const apiCollection = new ApiCollection();
-    for (const group of Object.keys(description)) {
-      apiCollection.set(group, await ApiGroup.init(description[group], group));
-    }
-    return apiCollection;
+    const promises: Promise<ApiGroup>[] = Object.entries(
+      description
+    ).map(([group, paths]) => ApiGroup.init(paths, group));
+
+    return new this(...(await Promise.all(promises)));
```
wjhsf, line 26:
```suggestion
    return this.from(await Promise.all(promises));
```
Array constructor can have weird behavior depending on the arguments. I don't think we'd ever run into anything unexpected, but I think that using `from` is clearer.

src/generate/apiCollection.test.ts
```diff
@@ -23,19 +23,17 @@ describe("Test ApiGroup class init", () => {
     const collection = await ApiCollection.init({
```
wjhsf, line 23: chai-as-promised is no longer used in this file and can be removed.

wjhsf, line 23: describe text needs to be updated

src/generate/apiCollection.test.ts
```diff
@@ -23,19 +23,17 @@ describe("Test ApiGroup class init", () => {
...
-    expect(collection.get("Group One").apis).to.not.be.empty;
-    expect(collection.get("Group One").apis[0].model).to.not.be.empty;
-    expect(collection.get("Group One").apis[0].name.original).to.equal(
-      "Shop API"
-    );
-    expect(collection.get("Group One").apis[0].path).to.be.equal(validRamlFile);
+    expect(collection.get("Group One")).to.not.be.empty;
+    expect(collection.get("Group One")[0].model).to.not.be.empty;
+    expect(collection.get("Group One")[0].name.original).to.equal("Shop API");
+    expect(collection.get("Group One")[0].path).to.be.equal(validRamlFile);
```
wjhsf, line 29:
```suggestion
    expect(collection.get("Group One")[0].path).to.equal(validRamlFile);
```
Also in a few other places...

src/generate/apiGroup.ts
```diff
@@ -5,19 +5,18 @@
...
  * A group of API objects. Common transformations of the group name are cached
  * for reference in templates and file paths.
  */
-export class ApiGroup {
-  apis: Api[];
+export class ApiGroup extends Array<Api> {
   name: Name;
 
-  constructor(name = "", apis: Api[] = []) {
+  constructor(name = "", ...apis: Api[]) {
```
wjhsf, line 17:
```suggestion
  constructor(name = "", apis: Api[] = []) {
```
I think it makes more sense for the arguments to be "a name and an array of all the APIs you want" rather than "all the APIs you want, except the first one is a name".

src/generate/apiGroup.ts
```diff
@@ -30,7 +29,16 @@ export class ApiGroup {
   static async init(apiSpecFilePaths: string[], name = ""): Promise<ApiGroup> {
```
wjhsf, line 29: Feels a bit weird to have `init(array, name)` and `constructor(name, ...array)`. Would be nice if they matched.

src/generate/apiCollection.ts
bredmond-sf: There are two overloads both for specifying size and doing `new ApiCollection(1)` appears to work fine in ts-node. I'm having problems with the 'from', though: `Property 'get' is missing in type 'ApiGroup[]' but required in type 'ApiCollection'.`

src/generate/apiCollection.test.ts
bredmond-sf: It's still used for the rejects tests at the bottom

bredmond-sf: good catch

src/generate/apiCollection.test.ts
bredmond-sf: Also good catches

src/generate/apiCollection.ts
```diff
@@ -10,6 +10,10 @@ import { ApiGroup } from "./apiGroup";
  * A collection of multiple groups of APIs.
  */
 export class ApiCollection extends Array<ApiGroup> {
+  constructor(groups: ApiGroup[]) {
```
wjhsf, line 13: The `ApiGroup` constructor has a default parameter. This should as well.

src/generate/api.ts
```diff
@@ -12,7 +12,7 @@ import {
   getAllDataTypes
 } from "../common/parser";
 
-import Name from "./name";
+import { Name } from "./name";
```
wjhsf, line 15: Doing a full review, noticed that `model` in the constructor is missing a type. Also, class definition can be simplified by using `public` in the constructor parameter list.

bredmond-sf:
I think all comments are addressed

PR #86, opened 2020-07-10
Gk/move template helpers/w 7768345

gkhunger-sf:
* Move all the handlebars helper function that help accessing information from an amf model from commerce-sdk to raml-toolkit
* Create a Handlebars environment and register all the handlebarsAmfHelpers functions
* Set Handlebars environment from handlebarsAmfHelpers as default in Template class 
* Assign specific return types to isTypeDefinition and isAdditionalPropertiesAllowed 
* Add documentation for the new handlebarsAmfHelpers module
* Move all the functions that don't need to be externally exposed in utils.ts
* Add tests for isTypeDefinition function
* Add AMF.init() to validateModel to fix all the unnecessary warnings


package.json
```diff
@@ -153,6 +155,7 @@
     "eslint-config-xo-typescript": "^0.19.0",
     "eslint-plugin-header": "^3.0.0",
     "eslint-plugin-prettier": "^3.1.0",
+    "handlebars-helpers": "^0.10.0",
```
bredmond-sf, line 158: This should be a dependency, not devDep

src/generate/handlebarsAmfHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+): boolean => {
+  return (
+    ramlTypeDefinition !== undefined &&
+    ramlTypeDefinition.closed !== undefined &&
+    ramlTypeDefinition.closed.value !== undefined &&
+    !ramlTypeDefinition.closed.value()
+  );
+};
+
+const helpers = {
```
bredmond-sf, line 194: I think the cleaner way to do this is just move the handlebars config into its own file. Then you can simply register everything in this file to it (and so can other people). I also prefer having a top-level package export of 'HandlebarsWithAmfHelpers' instead of it being nested inside the helpers export.

src/generate/handlebarsAmfHelpers.ts
```diff
@@ -0,0 +1,215 @@
...
+  getValue,
+  isAdditionalPropertiesAllowed,
+  isOptionalProperty,
+  isRequiredProperty,
+  isTypeDefinition
+};
+
+const handlebars = Handlebars.create();
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require("handlebars-helpers")({ handlebars }, ["string", "comparison"]);
```
bredmond-sf, line 210: Drop the "string" and "comparison" and just make all the helpers available

src/generate/handlebarsAmfHelpers.ts
bredmond-sf: Also, this appears to just be a list of keys with no values which means that nothing gets registered below :/

src/generate/index.ts
```diff
@@ -8,3 +8,6 @@ export { Api } from "./api";
 export { ApiGroup } from "./apiGroup";
 export { ApiCollection } from "./apiCollection";
 export { Template } from "./template";
+
+import * as handlebarsAmfHelpers from "./handlebarsAmfHelpers";
+export { handlebarsAmfHelpers };
```
bredmond-sf, line 13: I'm curious if we should we should also or instead this include a subpath export? https://nodejs.org/api/esm.html#esm_subpath_exports

package.json
gkhunger-sf: Making it a dependency makes snyk very unhappy. Let me look a little bit into it.

src/generate/handlebarsAmfHelpers.ts
gkhunger-sf: Will move the config to a separate file.

ES6 allows just using values if they have the same names as the keys - https://alligator.io/js/object-property-shorthand-es6/

bredmond-sf: You are right, I always forget that one. But my point is that you don't need the config at all.

test/generate/handlebarsAmfHelpersTestUtils.ts
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+"use strict";
+import { expect } from "chai";
+import _ from "lodash";
```
wjhsf, line 9:
```suggestion
```
lodash is not used in this file.

src/generate/index.ts
gkhunger-sf: @bredmond-sf I didn't exactly use subpath-exports but imported all the functions we want to export in index.ts and exporting everything there. Does it satisfy the concern you had or do you still want me to look into using subpath-exports?

src/generate/handlebarsConfig.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { getValue } from "./utils";
```
wjhsf, line 8: `getValue` is being used as a helper, so I think it should be exported from the helpers file so that it doesn't need special handling.

src/generate/index.ts
```diff
@@ -9,5 +9,31 @@ export { ApiGroup } from "./apiGroup";
...
+  getRequestPayloadType,
+  getReturnPayloadType,
+  getProperties,
+  isAdditionalPropertiesAllowed,
+  isOptionalProperty,
+  isRequiredProperty,
+  isTypeDefinition
+} from "./handlebarsAmfHelpers";
+import { getValue } from "./utils";
+const handlebarsAmfHelpers = {
```
wjhsf, line 25: Why is this being exported as an object instead of what you had before?

src/lint/index.ts
```diff
@@ -31,6 +32,7 @@ export default class LintCommand extends Command {
 
     const promises = [];
 
+    await AMF.init();
```
wjhsf, line 35: Maybe add a comment explaining why this is here?

src/generate/index.ts
```diff
@@ -8,3 +8,32 @@ export { Api } from "./api";
...
+  getPropertyDataType,
+  getRequestPayloadType,
+  getReturnPayloadType,
+  getProperties,
+  isAdditionalPropertiesAllowed,
+  isOptionalProperty,
+  isRequiredProperty,
+  isTypeDefinition
+} from "./handlebarsAmfHelpers";
+import { getValue } from "./utils";
```
bredmond-sf, line 24: Why not just move getValue to handlebarsAmfHelpers or at least just export it from there? This now gives us another place to update when adding a helper.

src/generate/index.ts
gkhunger-sf: I misinterpreted a comment and went on a completely different route with getValue and all this. Reverting it.

src/generate/handlebarsAmfHelpers.ts
```diff
@@ -0,0 +1,193 @@
...
+): string => {
+  return property && property.encodes
+    ? (property.encodes as model.domain.WebApi).servers[0].url.value()
+    : "";
+};
+
+/**
+ * Checks the node is a type definition.
+ *
+ * @param obj - The node to check
```
monkey-jeff, line 41: docstring doesn't match method definition

src/generate/handlebarsAmfHelpers.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @returns true if the node is a type definition, false if not
+ */
+export const isTypeDefinition = (
+  cdProperty: model.domain.DomainElement
```
monkey-jeff, line 46: What does cdProperty stand for?

src/generate/handlebarsAmfHelpersGetType.test.ts
```diff
@@ -0,0 +1,274 @@
+/*
```
monkey-jeff, line 1: Why are these tests split out into its own file?

src/generate/utils.ts
```diff
@@ -0,0 +1,272 @@
...
+export const OBJECT_DATA_TYPE = "object";
+export const PRIMITIVE_DATA_TYPE_MAP = {
+  "http://www.w3.org/2001/XMLSchema#string": "string",
+  "http://www.w3.org/2001/XMLSchema#integer": "number",
+  "http://www.w3.org/2001/XMLSchema#double": "number",
+  "http://www.w3.org/2001/XMLSchema#float": "number",
+  "http://www.w3.org/2001/XMLSchema#boolean": "boolean"
+};
+
+export const getPayloadResponses = (
```
monkey-jeff, line 21: No method documentation

src/generate/utils.ts
```diff
@@ -0,0 +1,272 @@
...
+  return okResponses;
+};
+
+/**
+ * Given a payload, extract the types.
+ *
+ * @param payload - Contains schema(s) from which to extract the type(s).
+ * @returns string representation of the datatypes in the payload
+ */
+export const extractTypeFromPayload = (
```
monkey-jeff, line 39: All other methods seem to have the nomenclature of `getFoosBar`  I kind of like `getBarFromFoo`  but this seems to be the only one saying 'extract'

src/generate/handlebarsAmfHelpers.ts
gkhunger-sf: This type was CustomDomainProperty earlier and I changed it to DomainElement later but forgot to change the name. Will change it.

src/generate/handlebarsAmfHelpersGetType.test.ts
gkhunger-sf: Otherwise the file becomes too big, bigger than our current limit of 500 lines.

PR #91, opened 2020-07-14
Add diff directories flag to diff command

wjhsf:
- New (sub)command: `raml-toolkit diff --dir old/api-config.json new/api-config.json` to replace commerce-sdk's `diffRamlFiles` gulp task.
- Added `npm run docs` script to cleanly update the README, because `oclif-dev` reads from the compiled code.


src/diff/diffDirectories.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Compares two arrays and returns their intersection and symmetric difference.
+ *
+ * @param leftArr - One of the arrays being compared
+ * @param rightArr - The other array being compared
+ *
+ * @returns An array each for all the common, exclusive to left and exclusive to
+ * right elements.
+ */
+function compareArrays<T>(
```
monkey-jeff, line 42: Hmm. Not sure about the name "compareArrays" but I am at a loss for a better name.

src/diff/diffDirectories.ts
```diff
@@ -0,0 +1,139 @@
...
+  leftArr: T[],
+  rightArr: T[]
+): {
+  leftOnly: T[];
+  rightOnly: T[];
+  common: T[];
+} {
+  const left = new Set(leftArr);
```
monkey-jeff, line 50: Do we want to eliminate duplicate entries? If left contains [1,1,3] and right contains [1,3] this would basically show everything to be in common when there is an extra 1 on the left side.

src/diff/diffDirectories.ts
```diff
@@ -0,0 +1,139 @@
...
+  const common = new Set<T>();
+  left.forEach(val => {
+    if (right.has(val)) {
+      left.delete(val);
+      right.delete(val);
+      common.add(val);
+    }
+  });
+  return {
+    leftOnly: [...left],
```
monkey-jeff, line 61: If not reducing to a set like I was talking about before I think that these should be renamed as well.  Such as "common, remainingLeft, remainingRight" or something like that.

src/diff/diffDirectories.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    const result: RamlDiff[] = await diffNewAndArchivedRamlFiles(
+      leftApiConfigFile,
+      rightApiConfigFile
+    );
+
+    require("fs-extra").writeJsonSync(path.join(__dirname, "tmp.json"), result);
+
+    expect(result.length).to.equal(4);
+    expect(diffRamlStub.callCount).to.equal(4);
+    expect(result[0].file).to.equal("api1/api1.raml");
```
monkey-jeff, line 63: Does order if not might want to change these tests to be order agnostic in the results.  Would make for a more stable test.

README.md
```diff
@@ -28,38 +28,48 @@ The npm installs the binaries as both `raml-toolkit` and `ramlint` and they can
...
-  evaluation against any ruleset.
+  --diff-only              Only show differences without evaluating a ruleset
+
+  --dir                    Find the differences for all files in two directories
+
+DESCRIPTION
+  This command has three modes: ruleset, diff-only, and directory.
+     Ruleset mode (default) compares two files and applies a ruleset to determine if any changes are breaking.
+     Diff-only mode compares two files to determine if there are any differences, without applying a ruleset.
+     Directory mode compares all the files in two directories and determines if there are any differences.
```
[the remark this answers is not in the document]
wjhsf: This story is just a migration. Adding functionality can be a follow-up story.

src/diff/diffDirectories.ts
wjhsf: The function is only ever called with the results of `listRamlsFromConfig`. Will that function ever return an array with duplicates?

https://github.com/SalesforceCommerceCloud/raml-toolkit/blob/70b0a165d72606b9ccb4ab2eba00e0a8441bda5e/src/diff/diffDirectories.ts#L26

monkey-jeff: Then we should more specifically name this method and describe what it does.  IT was written to seem like a generic array differ but it isn't that so we should be explicit.

README.md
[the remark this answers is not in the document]
wjhsf: Created a story for this. https://gus.my.salesforce.com/a07B0000008WGO9IAO

README.md
```diff
@@ -28,38 +28,48 @@ The npm installs the binaries as both `raml-toolkit` and `ramlint` and they can
 **Note:** Some commands require environment variables to be set. This can be done using a [.env file](https://www.npmjs.com/package/dotenv#rules) in your working directory (the directory from which you run `raml-toolkit`).
 
 <!-- commands -->
-- [`raml-toolkit diff APISPECBASEPATH APISPECNEWPATH`](#raml-toolkit-diff-apispecbasepath-apispecnewpath)
+- [`raml-toolkit diff BASE NEW`](#raml-toolkit-diff-base-new)
```
unandyala, line 31: I didn't mean literally base and new. I mean to have base and new in place of old and new, something like BASEAPI NEWAPI or BASEAPISPEC NEWAPISPEC

Sorry for not being clear

README.md
```diff
@@ -28,38 +28,48 @@ The npm installs the binaries as both `raml-toolkit` and `ramlint` and they can
...
-  APISPECBASEPATH  The base API spec file for the comparison
-  APISPECNEWPATH   The new version of the API spec for comparison against the base version
+  BASE  The base API spec file (ruleset / diff-only mode) or configuration (directory mode)
+  NEW   The new API spec file (ruleset / diff-only mode) or configuration (directory mode)
 
 OPTIONS
-  -h, --help             show CLI help
-  -r, --ruleset=ruleset  Path to ruleset to apply to diff
-  -v, --version          show CLI version
+  -h, --help               show CLI help
```
gkhunger-sf, line 48: Can you please make all the descriptions for options uppercase?

src/diff/index.ts
```diff
@@ -28,60 +39,96 @@ The ruleset flag is used to evaluate a custom ruleset in place of the default ru
...
+  protected async _saveOrLog(file: string, json: object): Promise<void> {
+    if (file) {
+      await fs.writeJson(file, json);
+    } else {
+      console.log(json);
+    }
+  }
 
-    let results: NodeDiff[];
+  protected async _diffDirs(
```
gkhunger-sf, line 89: Can you please add tsdoc for these new functions

README.md
monkey-jeff: Capitalized...not uppercase please.

gkhunger-sf: 😂 you know what I meant

src/diff/diffDirectories.ts
```diff
@@ -33,20 +33,20 @@ export function listRamlsFromConfig(configPath: string): string[] {
...
+ * Expects arrays to have no duplicate elements.
  *
  * @param leftArr - One of the arrays being compared
  * @param rightArr - The other array being compared
  *
  * @returns An array each for all the common, exclusive to left and exclusive to
  * right elements.
  */
-function compareArrays<T>(
+function getCommonAndUniqueElements<T>(
```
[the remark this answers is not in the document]
wjhsf: I want to avoid `Set` because that's a JS class, and not what we return.

README.md
wjhsf: Oh, this makes more sense. I saw the email notifications and thought @gkhunger-sf was having a conversation with himself.

Also, no. Those are oclif defaults. (Technically, yes, but I have an idea that's outside the scope of this PR.)

README.md
wjhsf: Because oclif makes the names caps, I think that BASE / NEW are the cleanest for readability.

PR #93, opened 2020-07-16
Add common flags for commands

wjhsf:
Added three common flags:
- `help`: the default help flag, but hidden, `-h` enabled, and description updated to have capital first letter
- `version`: the default version flag, but hidden, `-v` enabled, and description updated to have capital first letter
- `verbosity`: A flag that sets the log level, without requiring the command to have any special handling

First two are to appease @gkhunger-sf. Last one is one I made for #78. I ended up not including it in that PR because the logging situation is a bit complex. I think it's still worth using, though, at least until we address the disparate logging systems.


src/common/flags.ts
```diff
@@ -0,0 +1,93 @@
...
+          ", "
+        )}`
+      );
+    }
+    return logger.getLevel();
+  }
+});
+
+/**
+ * The oclif help flag builder, with char set to -h, description changed to
```
unandyala, line 42:
```
description changed to
 * start with a capital letter,
```

Is this required in function doc?

src/common/flags.ts
```diff
@@ -0,0 +1,93 @@
...
+    char: "h",
+    description: "Show CLI help",
+    hidden: true,
+    ...options
+  });
+}
+
+/**
+ * The oclif version flag builder, with char set to -v, description changed to
+ * start with a capital letter, and hidden by default.
```
unandyala, line 59: same here

src/common/flags.ts
wjhsf: oclif uses lower case descriptions. We use sentence case. @gkhunger-sf didn't like the inconsistency, so this fixes it.

dbarbato-sf:
As discussed offline with @wjhsf , let's drop the -v shortcut for version and assign -v to verbose, this makes the cli feel a bit more standard.

README.md
```diff
@@ -45,17 +45,15 @@ ARGUMENTS
   NEW   The new API spec file (ruleset / diff-only mode) or configuration (directory mode)
 
 OPTIONS
-  -h, --help               show CLI help
-  -o, --out-file=out-file  File to store the computed difference
+  -V, --verbosity=trace|debug|info|warn|error|silent  [default: info] Set the level of detail in the output
```
dbarbato-sf, line 48: can we also make this --verbose not --verbosity ?

PR #94, opened 2020-07-17
Rename handlebars helpers and its utils

gkhunger-sf:
Rename  handlebars helpers and its utils based on Jeff's quip doc - https://salesforce.quip.com/JNnDAivz5cyK

### DIFF
--- a/src/generate/handlebarsAmfHelpers.ts
+++ b/src/generate/handlebarsAmfHelpers.ts
@@ -192,9 +192,7 @@ export const isAdditionalPropertiesAllowed = (
   node: model.domain.NodeShape
 ): boolean => {
   return (
-    node !== undefined &&
-    node.closed !== undefined &&
-    node.closed.value !== undefined &&
+    node?.closed?.value &&
     !node.closed.value()
   );
 };
--- a/src/generate/handlebarsAmfHelpers.ts
+++ b/src/generate/handlebarsAmfHelpers.ts
@@ -6,10 +6,10 @@
  */
 
 import {
-  getPayloadResponses,
+  getResponsesFromPayload,
   getTypeFromPayload,
   getDataType,
-  getPayloadType,
+  getTypeFromShape,
   getFilteredProperties,
   getValue,
   DEFAULT_DATA_TYPE,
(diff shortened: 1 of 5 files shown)


src/generate/utils.ts
```diff
@@ -228,22 +249,23 @@ export const getValue = <T>(name: model.ValueField<T>): string => {
...
+ * @param node - An AMF node
+ * @param propertyFilter - predicate to be used to filter properties
  *
- * @param dtoTypeModel - AMF model of the dto
- * @param propertyFilter - function to filter properties based on certain criteria
  * @returns The filtered list of properties
  */
 export const getFilteredProperties = (
-  dtoTypeModel: model.domain.NodeShape | null | undefined,
+  node: model.domain.NodeShape | null | undefined,
```
wjhsf, line 260: Will `null` or `undefined` ever actually be passed to the function?

src/generate/handlebarsAmfHelpers.ts
```diff
@@ -77,12 +79,13 @@ export const getReturnPayloadType = (
...
+ * Get data type of a property.
+ *
+ * @param property - An AMF property
  *
- * @param property - instance of model.domain.PropertyShape
- * @returns data type if defined in the property otherwise returns a default type
+ * @returns data type, if defined in the property, the default type otherwise
  */
-export const getPropertyDataType = (
+export const getTypeFromProperty = (
```
monkey-jeff, line 88: How does a property differ from a parameter?

src/generate/handlebarsAmfHelpers.ts
```diff
@@ -174,19 +182,20 @@ export const isOptionalProperty = (
...
 ): boolean => {
   return (
-    ramlTypeDefinition !== undefined &&
-    ramlTypeDefinition.closed !== undefined &&
-    ramlTypeDefinition.closed.value !== undefined &&
-    !ramlTypeDefinition.closed.value()
+    node !== undefined &&
+    node.closed !== undefined &&
+    node.closed.value !== undefined &&
+    !node.closed.value()
```
wjhsf, line 198:
```suggestion
    node?.closed?.value &&
    !node.closed.value()
```

src/generate/utils.ts
gkhunger-sf: Probably not. I didn't know the reason it was added for so didn't bother updating it.

src/generate/handlebarsAmfHelpers.ts
gkhunger-sf: The only difference is the kind of AMF element they operate on.
model.domain.PropertyShape vs model.domain.Parameter

src/generate/handlebarsAmfHelpers.ts
gkhunger-sf: Reverted this change since it made one of the tests fail.

wjhsf:
```suggestion
    typeof node?.closed?.value === 'function' && !node.closed.value()
```
¯\\\_(ツ)_/¯

gkhunger-sf: I trusted you once, I am not doing it again 😛

PR #107, opened 2020-07-31
Renamed test to testResources

monkey-jeff:
Test was misleading.  It is now testResources.


.depcheckrc.yml
```diff
@@ -1,2 +1,3 @@
 ignores:
   - dotenv
+  - sort-package-json
```
wjhsf, line 3:
```suggestion

```
depcheck complained because you removed the script that was using this

package.json
```diff
@@ -22,20 +22,26 @@
   ],
   "scripts": {
     "clean": "rm -rf lib",
+    "build": "npm run compile",
     "compile": "tsc -b && echo Compiled",
     "depcheck": "depcheck",
     "docs": "npm run clean && npm run compile && oclif-dev readme",
-    "lint": "npm run lint:dev -- --quiet",
-    "lint:dev": "sort-package-json && eslint . --ext .ts --fix",
+    "lint": "eslint . --ext .ts --fix",
```
wjhsf, line 29:
```suggestion
    "lint": "sort-package-json && eslint . --ext .ts --fix",
```
This should not be removed.

PR #108, opened 2020-08-03
Organized exports for diff tool

unandyala:
1. Moved DiffCommand class from diff/index.ts to its own file - "diff/diffCommand.ts". Now diff/index.ts contains just the exports of diff tool
2. Diff tool functions are exported as "diff" from raml-toolkit
`export * as diff from "./diff";`

**Usage example:**
```
import { diff } from "@commerce-apps/raml-toolkit";
 const apiDifferencer = new diff.ApiDifferencer(
        path.join(__dirname, "../diffRamls/base/shopper-products/shopper-products.raml"),
        path.join(__dirname, "../diffRamls/new/shopper-products/shopper-products.raml")
    );
```

Note: I plan to do other modules in a similar way once this gets approved


src/index.ts
```diff
@@ -15,6 +15,9 @@ export { model } from "amf-client-js";
 
 export * from "./generate";
 
+// eslint-disable-next-line prettier/prettier
```
wjhsf, line 18: Would upgrading to a newer version of prettier make it not complain?

src/diff/diffCommand.ts
```diff
@@ -0,0 +1,184 @@
+/*
```
wjhsf, line 1: Consider renaming this file `diff/command.ts` for consistency with `download/command.ts`.

src/index.ts
unandyala: I didn't try.. I will try that in the followup PR

src/diff/diffCommand.ts
unandyala: thought about it.. I feel we might end up with too many command.ts files which is more confusing

wjhsf: They are in different directories, which are used in different contexts. I think that having a standard format is less confusing, not more.

unandyala: more confusing when searching for files by name. I am fine with renaming if team prefers

monkey-jeff: I agree with ujwala here. I don't love using the same name at every level as when you are searching for a file you now have to scroll through all those same files in different directories.  It can also easily lead to confusion.

I do agree we should be consistent, I like the idea that they are all prefixed by the command name.

PR #109, opened 2020-08-04
Exclude nodes without a type from the parsed result

unandyala:


src/diff/amfGraphDifferencer.ts
```diff
@@ -171,6 +171,15 @@ function parseGraphDelta(
...
 }
 
+/**
+ * Returns true if type of amf graph node is defined
+ * @param nodeType - Types of amf graph node
+ */
+function isNodeTypeDefined(nodeType: string[]): boolean {
+  //source map nodes may not have type defined
+  return nodeType != null && nodeType.length > 0;
+}
```
wjhsf, line 181:
```suggestion
function isNodeTypeDefined(nodeType: string[]): boolean {
  //source map nodes may not have type defined
  return nodeType.length > 0;
}
```
If `nodeType` is always `string[]`, then this is sufficient.

```suggestion
function isNodeTypeDefined(nodeType: string[] | undefined | null): nodeType is string[] {
  //source map nodes may not have type defined
  return nodeType?.length > 0;
}
```
If `nodeType` is not always `string[]`, then (1) the type should reflect that, (2) the return value should be a type predicate so typescript can use the information, and (3) you can use the safe navigation operator to be more concise.

src/diff/amfGraphDifferencer.ts
```diff
@@ -184,6 +193,12 @@ function parseNodePropDelta(
   nodeType: string[],
   nodeDelta: AmfGraphDeltaTypes.NodePropertyDelta
 ): NodeChanges {
+  if (!isNodeTypeDefined(nodeType)) {
+    ramlToolLogger.debug(
+      `Ignoring changes to the node with no type: ${nodeId}`
+    );
+    return undefined;
```
wjhsf, line 200:
```suggestion
    return;
```
`return` implicitly returns `undefined`.

src/diff/amfGraphDifferencer.ts
```diff
@@ -184,6 +193,12 @@ function parseNodePropDelta(
   nodeType: string[],
   nodeDelta: AmfGraphDeltaTypes.NodePropertyDelta
 ): NodeChanges {
```
wjhsf, line 195:
```suggestion
): NodeChanges | undefined {
```
The return type should reflect the returned values.

src/diff/amfGraphDifferencer.ts
```diff
@@ -269,13 +284,17 @@ export function parseNodeDelta(
   delta: AmfGraphDeltaTypes.ArrayElementDelta<AmfGraphTypes.Node>,
   deltaType: DeltaType
 ): NodeChanges {
-  let nodeChanges;
   const node = delta[0];
   const id = node[AmfGraphTypes.KEY_NODE_ID];
   const type = node[AmfGraphTypes.KEY_NODE_TYPE];
+  if (!isNodeTypeDefined(type)) {
+    ramlToolLogger.debug(`Ignoring changes to the node with no type: ${id}`);
+    return undefined;
```
wjhsf, line 292:
```suggestion
    return;
```
`return` implicitly returns `undefined`.

src/diff/amfGraphDifferencer.ts
```diff
@@ -269,13 +284,17 @@ export function parseNodeDelta(
   delta: AmfGraphDeltaTypes.ArrayElementDelta<AmfGraphTypes.Node>,
   deltaType: DeltaType
 ): NodeChanges {
```
wjhsf, line 286:
```suggestion
): NodeChanges | undefined {
```
The return type should reflect the returned values.

src/diff/amfGraphDifferencer.test.ts
```diff
@@ -68,7 +68,7 @@ describe("Validate flattened AMF graph", () => {
 describe("Changes to graph nodes", () => {
```
wjhsf, line 68: Chai recommends using `lengthOf` instead of `length`: https://www.chaijs.com/api/bdd/#method_lengthof

src/diff/amfGraphDifferencer.ts
unandyala: we do not allow implicit returns -  "noImplicitReturns" is set true

src/diff/amfGraphDifferencer.ts
```diff
@@ -171,6 +171,17 @@ function parseGraphDelta(
   return graphChanges;
 }
 
+/**
+ * Returns true if type of amf graph node is defined
+ * @param nodeType - Types of amf graph node
+ */
+function isNodeTypeDefined(
+  nodeType: string[] | undefined
+): nodeType is string[] {
```
kalagarraj, line 180: change return type to boolean

unandyala: https://www.typescriptlang.org/docs/handbook/advanced-types.html

PR #111, opened 2020-08-04
New Ignore rule for difftool to ignore the changes in RAML examples

kalagarraj:
Included changes
- New default rule to ignore example changes
- Test cases for default rule and node type structural changes
- Sorted package.json

### DIFF
--- a/resources/diff/rules/defaultRules.json
+++ b/resources/diff/rules/defaultRules.json
@@ -164,5 +164,24 @@
         "changedProperty": "core:version"
       }
     }
+  },
+  {
+    "name": "Rule to ignore changes in examples",
+    "conditions": {
+      "any": [
+        {
+          "fact": "diff",
+          "path": "$.type",
+          "operator": "contains",
+          "value": "apiContract:Example"
+        }
+      ]
+    },
+    "event": {
+      "type": "example-change",
+      "params": {
+        "category": "Ignored"
+      }
+    }
   }
 ]
--- a/resources/diff/rules/defaultRules.json
+++ b/resources/diff/rules/defaultRules.json
@@ -178,7 +178,7 @@
       ]
     },
     "event": {
-      "type": "example-change",
+      "type": "example-changed",
       "params": {
         "category": "Ignored"
       }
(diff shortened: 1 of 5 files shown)


src/diff/changes/nodeChanges.ts
```diff
@@ -14,6 +14,7 @@ import { RuleCategory } from "../ruleSet";
 export class NodeChanges {
   public added: { [key: string]: unknown };
   public removed: { [key: string]: unknown };
+  public ignored: { [key: string]: unknown };
```
unandyala, line 17: what is the purpose of ignored object? I don't see it being used anywhere

src/diff/defaultRules.test.ts
```diff
@@ -27,7 +31,7 @@ function verifyRule(nodeChanges: NodeChanges, rule: Rule): void {
       nodeChanges.added[rule.event.params.changedProperty]
     ];
   }
-  expect(categorizedChanges).to.have.length(1);
+  expect(categorizedChanges).to.have.length(countOfChanges);
```
unandyala, line 34: length seems to have deprecated. We should use lengthOf instead

src/diff/defaultRules.test.ts
```diff
@@ -120,3 +124,75 @@ describe("Version change", () => {
     verifyRule(changes[0], rule);
   });
 });
+
+describe("Examples change", () => {
+  it("does not apply ignore examples rule on a change that has no apiContract:Example change type", async () => {
```
unandyala, line 129: apiContract:Example is a node type not a change type

kalagarraj: Done

src/diff/defaultRules.test.ts
kalagarraj: done

src/diff/changes/nodeChanges.ts
kalagarraj: Removed

resources/diff/rules/defaultRules.json
```diff
@@ -164,5 +164,24 @@
...
+        {
+          "fact": "diff",
+          "path": "$.type",
+          "operator": "contains",
+          "value": "apiContract:Example"
+        }
+      ]
+    },
+    "event": {
+      "type": "example-change",
```
unandyala, line 181: change type to "example-changed" - just to be consistent with types in other rules

kalagarraj: Makes sense, done..

PR #118, opened 2020-08-13
Add rule to ignore description changes

unandyala:


wjhsf:
My PR changes the names of the rules. This new rule is consistent with the format I used, but "changed" should be capitalized. (Or maybe I should change the case of the rules in my PR 🤷)

unandyala:
> All the other rules are named like "Rule to ...". It would look good if they are all named the same way. Just this very minor thing, otherwise looks alright.

I changed because @wjhsf PR is changing it

PR #121, opened 2020-09-14
Ka/node fetch.w 8038871

kalagarraj:
Node fetch update to fix Snyk reported issue


unandyala (approved):
why didn't snyk create a PR for this? Did we disable it?

wjhsf (approved):

kalagarraj:
> why didn't snyk create a PR for this? Did we disable it?

Yes..Snyk suggested to fix this using CLI

PR #122, opened 2020-09-14
Ka/production versions.w 8038871

kalagarraj:
- Using v1 APIs for fetching Anypoint Asset
- Changed test setups for nock

Manual Tests
- npm pack
- linter works
- downloader works
- diff works

### DIFF
--- a/src/download/exchangeDownloader.ts
+++ b/src/download/exchangeDownloader.ts
@@ -23,8 +23,8 @@ import { extractFile } from "./exchangeDirectoryParser";
 
 const DEFAULT_DOWNLOAD_FOLDER = "download";
 const ANYPOINT_BASE_URI = "https://anypoint.mulesoft.com/exchange";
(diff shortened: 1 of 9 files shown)


src/download/exchangeDownloader.ts
```diff
@@ -149,7 +149,7 @@ export async function searchExchange(
   searchString: string
 ): Promise<RestApi[]> {
   return fetch(
-    `${ANYPOINT_BASE_URI}/assets?search=${searchString}&types=rest-api`,
+    `${ANYPOINT_API_URI_V2}/assets?search=${searchString}&types=rest-api`,
```
unandyala, line 152: why V2 here?

kalagarraj: It is just the variable name change...It was V2 always

monkey-jeff (commented):
I would feel better if we kept some negative tests with v2 endpoints to demonstrate how this does NOT work with V2.

gkhunger-sf (approved):

monkey-jeff (approved):

PR #124, opened 2020-09-17
Upgrade to AMF 4.3

wjhsf:
- Upgraded to AMF 4.3 and changed main lint function to not use build artifacts.
- Changed all but one validation to use property constraints instead of JS functions. Mostly copied / slightly tweaked [amf's test file](https://github.com/aml-org/amf/blob/master/amf-client/shared/src/test/resources/validations/profiles/mercury.yaml).
- Noticed that the SLAS profile has no tests, so I added a test. Since it's identical to the Mercury profile except for one validation, I just tested that the profile objects are the same, rather than duplicating the validation tests.
- ~Changed the resource tests to cache the mercury validation profile.~ Broke unit tests, so I reverted this.
- Added some tests to meet code coverage threshold.
- Deleted a duplicate test file.

There's one validation remaining that is still a function, but it needs to be converted to a SPARQL query instead of a property constraint. I'm talking with the AMF team about how to do that, but I figured I'd get this up in the meantime.


gkhunger-sf (dismissed):

bredmond-sf (dismissed):

gkhunger-sf (approved):

resources/lint/profiles/mercury.mercury-profile.test.ts
```diff
@@ -1,351 +0,0 @@
-/*
```
monkey-jeff, line 1: Why was this deleted?

resources/lint/profiles/slas.raml
```diff
@@ -177,23 +164,29 @@ validations:
...
       apiContract.path:
-        pattern: ^((\/[a-z\.][a-z0-9-]*[a-z0-9]+)|(\/{[^}]+}))+$
+        pattern: ^((\/[a-z][a-z0-9-]*[a-z0-9]+)|(\/\{[^}]+}))+$
   
   unique-display-name:
     message: displayName must be unique within an API
     targetClass: apiContract.Operation
+    # TODO: Convert this to not use functionConstraint
     functionConstraint:
       code: |
```
monkey-jeff, line 174: Is code still supported in 4.3?

monkey-jeff (commented):
A couple questions

monkey-jeff (approved):
Just noticed still using code in one spot. Is that still supported?

resources/lint/profiles/slas.raml
wjhsf: The [release notes for 4.2](https://github.com/aml-org/amf/releases/tag/4.2.0) say it's no longer supported, but the functionality still exists. I guess it's deprecated?

I decided to leave this one for later because I'm leaving the team and because it requires input from AMF, but the person who was helping is now on vacation. I've added a story so that someone can follow up later.

PR #125, opened 2020-09-18
Refactor directory diff to use api models

gkhunger-sf:
* Refactor diffDirectories to use the new api structure which replaces api-config.json with .metadata.json
* Load base and new api directories into ApiMetadata trees, and then traverse and compare both the trees
* Update diff directory tests to test the new logic

### DIFF
--- a/src/diff/diffDirectories.ts
+++ b/src/diff/diffDirectories.ts
@@ -20,8 +20,8 @@ import { loadApiDirectory, ApiMetadata, ApiModel } from "../generate";
  * @param changes - ApiCollectionChanges object to contain all the changes
  */
 export async function diffApiModels(
-  leftApi: ApiModel,
-  rightApi: ApiModel,
+  leftApi: ApiModel | undefined,
+  rightApi: ApiModel | undefined,
   changes: ApiCollectionChanges
 ): Promise<void> {
   if (leftApi && rightApi) {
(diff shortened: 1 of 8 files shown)


wjhsf (approved):

src/diff/diffDirectories.ts
```diff
@@ -9,125 +9,99 @@ import path from "path";
...
-    } catch (error) {
-      ramlToolLogger.error(`Diff operation for '${raml}' failed:`, error);
-      apiCollectionChanges.errored[raml] = error.message;
-    }
+  // If either of the nodes passed is an ApiModel object, then we have
+  // traversed that tree branch entirely and can pass the APIs for diff to
+  // be run upon
+  if (leftNode instanceof ApiModel || rightNode instanceof ApiModel) {
+    return diffApiModels(leftNode as ApiModel, rightNode as ApiModel, changes);
+  }
```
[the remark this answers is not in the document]
gkhunger-sf: This is to determine whether we have traversed through the directories and have reached the file level. Once that happens block will be executed on each file pair to find differences.

wjhsf (dismissed):

src/diff/diffDirectories.ts
```diff
@@ -9,125 +9,102 @@ import path from "path";
...
+ * @param newDir - Newly downloaded APIs
  *
- * @returns An array of RamlDiff objects containing differences between all the
- * old and new RAML files
+ * @returns An ApiCollectionChanges object containing all the changes
  */
 export async function diffRamlDirectories(
-  baseConfigFile: string,
-  newConfigFile: string
+  baseDir: string,
```
unandyala, line 99: Does this need a change in readme as the input is a directory instead of a config file?

gkhunger-sf: Yes, it does. I'll make the changes.

unandyala (approved):

wjhsf (approved):

PR #126, opened 2020-09-22
Add validation for input paths and error handling around amf parsing function

gkhunger-sf:
Also, add try/catch around amf function calls


bredmond-sf (approved):

src/common/amfParser.ts
```diff
@@ -38,8 +38,17 @@ export async function parseRamlFile(
...
   const parser = new amf.Raml10Parser(ccEnvironment);
+  let document: model.document.Document;
 
-  return parser.parseFileAsync(fileUri) as Promise<model.document.Document>;
+  try {
+    document = (await parser.parseFileAsync(
+      fileUri
+    )) as model.document.Document;
+  } catch (err) {
+    throw new Error(err.Yw);
```
wjhsf, line 48:
```suggestion
    // AMF throws a custom object instead of an instance of Error, so we need to
    // clean it up and re-throw it to make it more useful. AMF does not provide
    // any helpful way to access the underlying error (e.g. the ENOENT thrown by
    // fs.readFile), so the easiest way seems to be to convert the AMF error to
    // a string and then work with that string.

    if (err instanceof Error) {
      // Normal error, don't need to modify to rethrow
      throw err;
    }

    const message = `${err.message ?? err}`;
    // AMF error, message starts with amf name that we don't care about
    const cleaned = new Error(message.replace(/^amf\S+? /, "")) as Error & {
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      amfError: Record<string, any>;
    };
    cleaned.amfError = err;
    throw cleaned;
```
`err.Yw` is a build artifact, so #124 will break it. The suggested change is basically what I did in #124 to avoid using build artifacts.

src/diff/diffCommand.ts
```diff
@@ -192,12 +192,20 @@ Exit statuses:
...
   async run(): Promise<void> {
     const { args, flags } = this.parse(DiffCommand);
+    const baseApis = args.base;
+    const newApis = args.new;
+    if (!fs.pathExistsSync(baseApis)) {
+      this.error(`File or directory not found: ${baseApis}`);
+    }
+    if (!fs.pathExistsSync(newApis)) {
+      this.error(`File or directory not found: ${newApis}`);
+    }
```
wjhsf, line 202:
```suggestion
    if (!await fs.pathExists(baseApis)) {
      this.error(`File or directory not found: ${baseApis}`);
    }
    if (!await fs.pathExists(newApis)) {
      this.error(`File or directory not found: ${newApis}`);
    }
```
Using sync methods when you're already in an async function always seems silly to me.

src/diff/diffCommand.test.ts
```diff
@@ -111,6 +111,13 @@ const operationRemovedRuleset = createTempFile(
 );
 
 describe("raml-toolkit cli diff command", () => {
+  let baseDir: DirResult;
```
wjhsf, line 114: Now that we have better output for errors, maybe adjust tests to check that output as well, instead of just exit code.

src/lint/lint.ts
```diff
@@ -34,7 +34,13 @@ export async function validateCustom(
     // An unexpected error was generated, throw a clean version of it.
     throw new Error(message);
   }
-  const report = await Core.validate(amfModel, profileName, MessageStyles.RAML);
+  let report: client.validate.ValidationReport;
+  try {
+    report = await Core.validate(amfModel, profileName, MessageStyles.RAML);
+  } catch (err) {
+    throw new Error(err.Yw);
```
wjhsf, line 41: Same comment as before, but I change this in #124, so don't worry about it.

wjhsf (changes requested):
Requesting changes because of use of AMF build artifacts.

monkey-jeff (dismissed):

wjhsf (approved):

bredmond-sf (approved):

PR #133, opened 2020-11-04
Add retry option to fetch

gkhunger-sf:
- Replace "node-fetch" with "make-fetch-happen" which allows passing in retry options
- downloadApi function now throws error when even after retrying it fails to download an api. Earlier it logged the error when a request was unsuccessful.

### DIFF
--- a/src/download/exchangeDownloader.ts
+++ b/src/download/exchangeDownloader.ts
@@ -5,8 +5,9 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import _ from "lodash";
+import fetch from "make-fetch-happen";
 import fs from "fs-extra";
-import fetch from "node-fetch";
 import path from "path";
 
 import { getBearer } from "./bearerToken";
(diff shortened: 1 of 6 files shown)


src/download/exchangeDownloader.ts
```diff
@@ -20,12 +21,40 @@ import {
...
+ * @returns a promise with the raw response returned by the server
+ */
+export async function runFetch(
+  url: string,
+  options: { [key: string]: any } = {}
+): Promise<any> {
+  options.retry = _.merge({}, retryOptions, options.retry);
+  options = _.merge(options, {
+    onRetry() {
+      ramlToolLogger.info(
```
unandyala, line 49: Is there a way to log the error before retry or does it log by default?

unandyala (approved):

src/download/exchangeDownloader.ts
gkhunger-sf: I don't see anything like that available with make-fetch-happen

bredmond-sf (approved):

PR #136, opened 2020-11-12
Tests no longer print to console

gkhunger-sf:
* Add a setup file that contains mocha root hooks that stub console.log and ramlToolLogger.info methods
* Update mocha to 8.2.1
* Extract mocha, nyc and eslint configs to their own separate files.

### DIFF
--- a/src/lint/lint.test.ts
+++ b/src/lint/lint.test.ts
@@ -24,19 +24,20 @@ before(() => {
 });
 
 describe("#printResults", () => {
-  let consoleSpy: Sinon.SinonSpy;
+  let consoleStub: Sinon.SinonSpy;
 
   beforeEach(() => {
-    consoleSpy = sinon.spy(console, "log");
+    sinon.restore();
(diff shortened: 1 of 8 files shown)


bredmond-sf (approved):

src/lint/lint.test.ts
```diff
@@ -24,27 +24,28 @@ before(() => {
 });
 
 describe("#printResults", () => {
-  let consoleSpy: Sinon.SinonSpy;
+  let consoleStub: Sinon.SinonSpy;
 
   beforeEach(() => {
-    consoleSpy = sinon.spy(console, "log");
+    sinon.restore();
```
unandyala, line 30: Is this added to override/restore the stubbing done in setup.ts? Just trying to understand in the future if we want to stub/spy the logger in a specific test, we have to first restore it. Is that correct?

gkhunger-sf: correct

unandyala (approved):

PR #166, opened 2021-02-26
Ja/w 8930371.refactor to libraries

monkey-jeff:
This refactors our profiles to utilize the feature of libraries from AMF.  

This has the nice side effect of greatly reducing test times.


src/lint/lint.ts
```diff
@@ -36,13 +37,12 @@ export async function validateCustom(
     // fs.readFile), so the easiest way seems to be to convert the AMF error to
     // a string and then work with that string.
 
-    const message = `${err.message ?? err}`;
-    if (message.includes("no such file or directory")) {
```
monkey-jeff, line 40: This code was removed because it was swallowing errors related to a missing library instead of just a missing profile.

gkhunger-sf (dismissed):

src/lint/lint.test.ts
```diff
@@ -126,3 +136,31 @@ describe("#validateCustom", () => {
     stub.restore();
   });
 });
+
+describe("#validateModel", () => {
+  let testModel: model.document.Document;
+
+  beforeEach(async () => {
+    await AMF.init();
```
wjhsf, line 144: Doesn't this only need to be done once, rather than before every test?

testResources/testUtils.ts
```diff
@@ -14,6 +14,37 @@ import { expect } from "chai";
...
+};
+
+export function generateValidationRules(lib: string, rules: string[]): {} {
+  const myValidations: { [key: string]: string } = {};
+  rules.forEach((rule) => (myValidations[rule] = `${lib}.${rule}`));
+  return {
+    violation: rules,
+    validations: myValidations,
+  };
+}
```
wjhsf, line 36:
```suggestion
export function generateValidationRules<T extends string>(lib: string, rules: T[]): { violation: T[]; validations: { [key: T]: string } } {
  const myValidations: { [key: T]: string } = {};
  rules.forEach((rule) => { myValidations[rule] = `${lib}.${rule}`; });
  return {
    violation: rules,
    validations: myValidations,
  };
}
```
The type `{}` means an object with no properties, so trying to access `generateValidationRules(lib, rules).violation` will result in a type error.

wjhsf (approved):
The profiles use `.raml` as their file extension, even though they're validation profiles, not RAML. Should they be changed to `.yaml` like the libraries?

monkey-jeff:
> The profiles use `.raml` as their file extension, even though they're validation profiles, not RAML. Should they be changed to `.yaml` like the libraries?


This is somethign we should do but I didn't want to try and fix this as part of this story

src/lint/lint.test.ts
monkey-jeff: Perhaps.  But I didn't want to change this as part of this PR.

wjhsf (approved):

gkhunger-sf (approved):

PR #188, opened 2021-07-07
Synk updates

unandyala:


package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@oclif/command": "1.8.0",
     "@oclif/config": "1.17.0",
-    "amf-client-js": "4.7.0",
+    "amf-client-js": "4.7.1",
```
wjhsf, line 55: Wasn't there a release that broke stuff? I don't remember what version it was.

unandyala: I believe it is 4.7.3

unandyala: confirmed with @gkhunger-sf that it was a compilation error with 4.7.3.  With 4.7.1 the build is successful and able to lint the raml successfully

wjhsf (approved):

echessman (approved):

PR #217, opened 2023-01-05
chore: replace circle CI with GH actions

joeluong-sfcc:
This PR replaces circle CI functionality with github actions. Two workflows have been added:

- test.yml: Runs unit tests on linux runners against node versions 10, 12, 14, and 16
- publish.yml: Publishes RAML toolkit to npm when a github release is created
    - NPM tokens will have to rotated every year 

Created a new PR to remove circle CI status checks, but then realized that was configurable in settings


unandyala (dismissed):
Do we need to update our release doc?

.github/workflows/test.yml
```diff
@@ -0,0 +1,32 @@
...
+on:
+  push:
+    branches-ignore: [main]
+  workflow_dispatch:
+
+jobs:
+  linux-unit-tests:
+    strategy:
+      matrix:
+        node: [10, 12, 14, 16]
```
bredmond-sf, line 12: I don't see any reason to test this back to 10 and 12 is questionable https://endoflife.date/nodejs
Do we have a story to support 18 which is the current LTS?

joeluong-sfcc: Very good point, I'll create a follow up story to figure out node versioning support

p-biyani (approved):

unandyala (approved):

redmondb1 (approved):

PR #221, opened 2023-04-27
feat: add custom field for class name @W-13013140@

joeluong-sfcc:
This PR exposes a custom field, `x-salesforce-sdk-class-name` within the RAML file that is used for the class name for SDK generation. Previously, the class name in the SDK was defined by using the title field. Developers and Technical writers would update the title in the RAML files as they are also used for client facing documentation, leading to unintended downstream impacts onto the SDKs. To help combat these breaking changes, `x-salesforce-sdk-class-name` can be defined explicitly in the RAML file to be used for SDK class name instead of title. If the custom field does not exist, the SDK will use the title by default.

This also allows Isomorphic SDK hot fix to be reverted once the SDK is updated to consume the new RAML toolkit and the Shopper Context RAML is updated to use the custom field: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/120

The main logic for this change is located in `src/generate/apiModel.ts`.

### DIFF
--- a/src/generate/apiModel.test.ts
+++ b/src/generate/apiModel.test.ts
@@ -129,11 +129,11 @@ describe("ApiModel tests", () => {
     const api = new ApiModel("CUSTOM", path.dirname(customFieldRamlFile));
     await api.init();
     expect(api.model.encodes.customDomainProperties.length).to.be.equal(2);
-    // x-salesforce-sdk-name: Custom Shop API Name
+    // x-salesforce-sdk-class-name: Custom Shop API Name
     expect(
       api.model.encodes.customDomainProperties[0].name.toString()
     ).to.be.equal(CUSTOM_NAME_FIELD);
-    // x-salesforce-sdk-name: Custom Shop API Name 2 (not used)
+    // x-salesforce-sdk-class-name: Custom Shop API Name 2 (not used)
     expect(
       api.model.encodes.customDomainProperties[1].name.toString()
     ).to.be.equal(CUSTOM_NAME_FIELD);
--- a/src/generate/apiModel.test.ts
+++ b/src/generate/apiModel.test.ts
@@ -13,7 +13,7 @@ import chaiFs from "chai-fs";
 import fs from "fs-extra";
 
 import { ApiModel } from "./";
-import { Name } from "../common/structures/name";
+import { Name, CUSTOM_NAME_FIELD } from "../common/structures/name";
 import tmp from "tmp";
 import sinon from "sinon";
 
@@ -27,6 +27,16 @@ const validRamlFile = path.join(
 const invalidRamlFile = path.join(
   __dirname,
   "../../testResources/raml/invalid/search-invalid.raml"
+);
+
+const customFieldRamlFile = path.join(
+  __dirname,
+  "../../testResources/raml/custom-field/custom-field.raml"
+);
+
+const customFieldInvalidRamlFile = path.join(
+  __dirname,
+  "../../testResources/raml/custom-field-invalid/custom-field-invalid.raml"
 );
 
 const handlebarTemplate = path.join(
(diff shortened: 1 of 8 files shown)


src/generate/apiModel.test.ts
```diff
@@ -102,6 +112,46 @@ describe("ApiModel tests", () => {
       "Cannot update the name before the model is loaded"
     );
   });
+
+  it("uses title for class name if custom field is not set", async () => {
+    const api = new ApiModel("VALID", path.dirname(validRamlFile));
+    await api.init();
+    expect(api.name).to.deep.equal(new Name("Shop API"));
```
tarcang, line 119: Why did you add 'deep' here, is it necessary?

p-biyani (approved):

src/generate/apiModel.ts
```diff
@@ -72,9 +72,23 @@ export class ApiModel extends ApiMetadata {
...
+
+    // If user defines custom class name under `x-salesforce-sdk-class-name`, use that instead
+    const customClassNameArr = this.model.encodes.customDomainProperties.filter(
+      (customProperty) => customProperty.name.toString() === CUSTOM_NAME_FIELD
     );
+    if (customClassNameArr.length > 0) {
+      // NOTE: there can be multiple instances of `x-salesforce-sdk-class-name` defined, we take the value of the first one
+      const customClassName = customClassNameArr[0].extension
+        .toString()
+        .split("=")[1];
```
tarcang, line 87: What is the purpose of split('=') here? How is the '=' symbol used while getting the custom title?

unandyala (approved):
we should post in cc-mercury-dev-internal channel so that they can start using this field

src/generate/apiModel.ts
```diff
@@ -72,9 +72,23 @@ export class ApiModel extends ApiMetadata {
...
       throw new Error("Cannot update the name before the model is loaded");
     }
-    this.name = new Name(
-      (this.model.encodes as model.domain.WebApi)?.name.value()
+
+    // class name is defaulted to title
+    let className = (this.model.encodes as model.domain.WebApi)?.name.value();
+
+    // If user defines custom class name under `x-salesforce-sdk-class-name`, use that instead
+    const customClassNameArr = this.model.encodes.customDomainProperties.filter(
```
unandyala, line 80: [NIT] extract the name resolution into its own function/method

testResources/raml/custom-field/custom-field.raml
```diff
@@ -0,0 +1,10 @@
+#%RAML 1.0
+
+title: Shop API
+x-salesforce-sdk-class-name: Custom Shop API Name
```
bredmond-sf, line 4: In RAML this would be `(salesforce-sdk-class-name)`. https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#annotations
as opposed to https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#x-other

and what if you followed our generator convention of:
```
(commerce-sdk):
  class-name: Custom Shop API Name
```

bredmond-sf (changes requested):

src/generate/apiModel.ts
joeluong-sfcc: There was a better way to get the custom field so I updated this part, but originally the string we were splitting on looked like: `scalar_1:http://www.w3.org/2001/XMLSchema#string=Custom Shop API title`, and we'd split on the equals sign to grab the `Custom Shop API title`

testResources/raml/custom-field/custom-field.raml
joeluong-sfcc: Done! You can see this in the example test file `testResources/raml/custom-field/custom-field.raml`

src/generate/apiModel.test.ts
joeluong-sfcc: I believe that `deep` recursively goes through the object to ensure that all properties are equal and structure of the object is the same. I just tested `expect(api.name).to.equal(new Name("Shop API"));` without deep and the unit test fails

unandyala (approved):

PR #224, opened 2024-05-28
@W-15675290 - Update getVersion removing deployment flag

adamraya:
**Problem**
The RAML spec is no longer deployed to MuleSoft, meaning that we no longer have the deployment information in Anypoint Exchange.

**Solution**
The PR addresses this issue by updating the logic to always retrieve the API version from the RAML `asset.version` value. This ensures that the latest RAML specification published to Anypoint Exchange is always used.
- The `download` command's options `--deployment` and `--deployment-regex-flags` are now deprecated and will be ignored.
- The `download` methods `search` and `getVersionByDeployment` have the `deployment` argument optional and will be ignored.


src/download/exchangeDownloader.ts
```diff
@@ -206,39 +206,50 @@ export async function searchExchange(
...
-    ) {
-      version = instance.version;
-    }
-  });
-  // If no instance matched the intended deployment get the version info
-  // from the fetched asset.
-  return version || asset.version;
+
+  if (!asset.version) {
+    ramlToolLogger.error(`${logPrefix} asset.version is missing`);
```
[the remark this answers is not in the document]
adamraya: Done!

src/download/downloadCommand.ts
```diff
@@ -47,10 +46,13 @@ export class DownloadCommand extends Command {
     }
     const { flags } = this.parse(DownloadCommand);
 
-    const apis = await download.search(
-      flags.search,
-      new RegExp(flags.deployment, flags["deployment-regex-flags"])
-    );
+    if (flags.deployment || flags["deployment-regex-flags"]) {
+      this.warn(
+        "The options 'deployment' and 'deployment-regex-flags' are deprecated and currently non-functional due to changes in the RAML spec. These options will be removed in the next major version."
```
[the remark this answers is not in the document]
adamraya: I agree that the message is more accurate I updated the copy.

src/download/exchangeDownloader.ts
```diff
@@ -206,39 +206,50 @@ export async function searchExchange(
...
+      accessToken,
+      `${restApi.groupId}/${restApi.assetId}`
+    );
+  } catch (error) {
+    ramlToolLogger.error(`${logPrefix} Error fetching asset:`, error);
+    return;
+  }
+
   if (!asset) {
+    ramlToolLogger.log(`${logPrefix} No asset found, returning`);
```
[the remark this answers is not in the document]
adamraya: Done

src/download/exchangeDownloader.ts
```diff
@@ -206,39 +206,54 @@ export async function searchExchange(
...
-  deployment: RegExp
+  deployment?: RegExp
 ): Promise<void | string> {
-  const asset = await getAsset(
-    accessToken,
-    `${restApi.groupId}/${restApi.assetId}`
-  );
+  if (deployment) {
+    ramlToolLogger.warn(
+      "Warning: The 'deployment' argument is deprecated and will be ignored due to changes in the RAML spec. This argument will be removed in the next major version."
```
unandyala, line 224: warning is not updated/modified?

adamraya: Good catch. I updated the message there as well.

unandyala (approved):

unandyala (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## 0.5.12
+
+* Fix bug by deprecating `deployment` argument from exchange downloader `search` command the argument is now optional and ignored.
+* Fix bug by deprecating `deployment` argument from `getVersionByDeployment` the argument is now optional and ignored.
```
joeluong-sfcc, line 6: Is there a specific reason we classified this as a patch fix instead of a major bump? I'm not sure if this can be considered a bug fix as this seems more of an implementation change where we no longer consider the deployed environment, which is breaking.

If we do a major version bump we could just remove the deprecated arguments alongside the warning messages.

unandyala: The only reason to avoid major version bump is to avoid supporting multiple versions. Since the main/only consumers of the raml toolkit is sdks, I think it is fine if we prefer major version bump

adamraya: The raml-toolkit is currently broken, we get the API specs from the latest version deployed to Anypoint Exchange, not the latest API specs. The reasoning for avoiding a new major version is that the update does not introduce new features but what we can consider as addressing an existing bug. The `deployment` argument was broken due to changes in the RAML spec, and the change fixes this by always returning the latest deployed version.

If we choose to do a minor version bump we maintain the API signature, we ensure backward compatibility and provide a smooth transition without forcing code updates in other repositories removing the deployment argument.
It's true that the result changes, but the change can be considered that is correcting the output rather than breaking the API, and we inform users by adding deprecation warnings.  I think makes sense to classify this as a bugfix. But I'm also open to changing the code to a major version bump.

joeluong-sfcc (dismissed):

src/download/exchangeDownloader.ts
```diff
@@ -206,39 +208,52 @@ export async function searchExchange(
...
+  if (!asset.version) {
+    ramlToolLogger.error(
+      `${logPrefix} The rest API ${restApi.assetId} is missing the asset.version`
+    );
+    return;
+  }
+
+  const finalVersion = asset.version;
+
+  return finalVersion;
```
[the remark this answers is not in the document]
adamraya: Done

unandyala (approved):

joeluong-sfcc (approved):

PR #228, opened 2025-05-15
@W-17701650@ Add support for OAS

vcua-mobify:
This PR introduces support for OAS templates by:

* Introducing `openapi-generator` to facilitate SDK generation from OAS as well as template initialization
* Introducing `oasdiff` to handle checking for API spec changes
* Updates to the download command so that it is able to download either OAS or RAML files

Notes: 
This is not intended to be a breaking change. All raml facilities will continue to work as-is.

The `lint` command is out of scope for this change.  `raml-toolkit lint` will only support RAML files.

### DIFF
--- a/src/download/exchangeDownloader.ts
+++ b/src/download/exchangeDownloader.ts
@@ -61,11 +61,12 @@ export async function runFetch(
 
 export async function downloadRestApi(
   restApi: RestApi,
-  destinationFolder: string = DEFAULT_DOWNLOAD_FOLDER
+  destinationFolder: string = DEFAULT_DOWNLOAD_FOLDER,
+  isOas: boolean = false
 ): Promise<void> {
   if (!restApi.id) {
     ramlToolLogger.warn(
-      `Failed to download '${restApi.name}' RAML as Fat RAML download information is missing.`,
+      `Failed to download '${restApi.name}' RAML / OAS as download information is missing.`,
       `Please download it manually from ${ANYPOINT_BASE_URI}/${restApi.groupId}/${restApi.assetId} and update the relevant details in apis/api-config.json`
     );
     return;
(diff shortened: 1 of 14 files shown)


unandyala (commented):
The way I am understanding is..
 we really need only the exchangeDownloader. Other tools are only wrappers around the openapi tools

Why do we need the wrapper?
Why can't we directly call the open api command? This seems to be an unnecessary additional layer.

My suggestion is we should directly call open api tools in the SDKs and just use the downloader from raml toolkit. This way it is easier to deprecate the raml-toolkit in future

src/download/exchangeDownloader.ts
```diff
@@ -119,6 +127,12 @@ function mapCategories(categories: RawCategories[]): Categories {
 
 function getFileByClassifier(files: FileInfo[], classifier: string): FileInfo {
   const found = files.find((file) => file.classifier === classifier);
+
+  // Not all API files in anypoint exchange have an associated fat-as classifier. so
```
unandyala, line 131:
```suggestion
  // Not all API files in anypoint exchange have an associated fat-oas classifier. so
```

unandyala (dismissed):
Left few minor comments

src/download/exchangeDownloader.ts
```diff
@@ -119,6 +127,12 @@ function mapCategories(categories: RawCategories[]): Categories {
...
+  // we return null here for those cases
+  if (!found) {
```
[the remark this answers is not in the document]
vcua-mobify: This is because we run the same getFileByClassifier for both oas and raml files: https://github.com/SalesforceCommerceCloud/raml-toolkit/blob/oas-support/src/download/exchangeDownloader.ts#L160

But for the raml files up on anypoint exchange, they won't have a `fat-oas` so we early exit rather than trying to read properties from null.

unandyala (approved):

README.md
```diff
@@ -26,13 +26,65 @@ npm install -g @commerce-apps/raml-toolkit
...
+  -c, --configFile=config_file                    A yaml file containing configurations for openapi-generator-cli. A default configuration will be used if this is not provided.
+
+  -g, --generator                                 The generator. Defaults to `typescript-fetch`
+
+  --skipValidateSpec                               Generate API without validating the spec. Defaults to false.
+
+
+#### `raml-toolkit diff BASE NEW -s oas`
+
+**Note:** This command requires that oasdiff is. installed. Install oasdiff according to https://github.com/oasdiff/oasdiff#installation
```
alexvuong, line 65: Nit: Let's move this to the top of the file to make it more visible before users run into an error trying to using oas

alexvuong:
Tested these changes locally with `commerce-sdk-isomorphic`. Confirmed I can the new oas is running as expected

alexvuong (approved):

alexvuong (approved):

unandyala (approved):

PR #231, opened 2025-06-04
@W-18706176: Changes to download latest release version of the API

unandyala:
@W-18706176 @W-18763414

1. Get latest version form all the version groups (V1, V2) of an API. This is to suuport Shopper Baskets V1 and V2
2. Remove the deprecated deployment flag
3. Added tests exchangeDirectoryParser to meet the coverage threshold
4. Spilt the test for exchangeDownloader.ts as there is a warning on the file size
5. All the code is behind isOas flag

Testing

1. download command
`node bin/run download --dest=raml-test --search='category:Visibility = "External"' --log-level=debug `

2. ran updateApis script in commerce-isomorphic-sdk

### DIFF
--- a/src/download/exchangeDownloader.ts
+++ b/src/download/exchangeDownloader.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, salesforce.com, inc.
+ * Copyright (c) 2025, salesforce.com, inc.
  * All rights reserved.
  * SPDX-License-Identifier: BSD-3-Clause
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
@@ -11,6 +11,7 @@ import fs from "fs-extra";
 import path from "path";
 
 import { getBearer } from "./bearerToken";
+import { removeVersionSpecificInformation } from "./exchangeTools";
 import {
   RawRestApi,
   RestApi,
@@ -26,6 +27,9 @@ export const DEFAULT_DOWNLOAD_FOLDER = "download";
 const ANYPOINT_BASE_URI = "https://anypoint.mulesoft.com/exchange";
 const ANYPOINT_API_URI_V1 = `${ANYPOINT_BASE_URI}/api/v1`;
 const ANYPOINT_API_URI_V2 = `${ANYPOINT_BASE_URI}/api/v2`;
+const DEPLOYMENT_DEPRECATION_WARNING =
+  "The 'deployment' argument is deprecated. The latest RAML specification that is published to Anypoint Exchange will be downloaded always.";
+
 // Only allows MAJOR.MINOR.PATCH (no suffixes). see https://semver.org/
 const releaseSemverRegex = /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$/;
 /**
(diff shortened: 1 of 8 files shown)


vcua-mobify:
Thanks for this @unandyala . The changes have ensured that I download the latest stable oas specs.

I wonder if we want to have a toggle where we can disable the check? Just wondering if pulling a snapshot spec for testing purposes is a valid use case we have to account for.

unandyala:
> Thanks for this @unandyala . The changes have ensured that I download the latest stable oas specs.
> 
> I wonder if we want to have a toggle where we can disable the check? Just wondering if pulling a snapshot spec for testing purposes is a valid use case we have to account for.

I don't think we need other versions. There will be lot of them and they are mainly developer versions. They may not even have backend implementation

src/download/exchangeDownloader.ts
```diff
@@ -259,15 +256,40 @@ export async function getVersionByDeployment(
...
+  versions: Array<{ version: string }>;
+}): void | string {
+  if (!versionGroup.versions || versionGroup.versions.length === 0) {
+    return;
+  }
+  const releaseAssetVersions = versionGroup.versions.filter((version) => {
+    return releaseSemverRegex.test(version.version);
+  });
+  // Sort versions and get the latest
+  return releaseAssetVersions.sort((instanceA, instanceB) => {
```
joeluong-sfcc, line 285: This function is a nice solution, it feels like a very clean way to get the latest version, straightforward to understand

src/download/exchangeDownloader.ts
```diff
@@ -297,28 +319,28 @@ export async function getSpecificApi(
...
  */
-export async function search(
-  query: string,
-  deployment?: RegExp
-): Promise<RestApi[]> {
-  if (deployment) {
-    ramlToolLogger.warn(DEPLOYMENT_DEPRECATION_WARNING);
-  }
-
+export async function search(query: string): Promise<RestApi[]> {
```
joeluong-sfcc, line 325: We're planning to search based on the API `SDK Type` and `Visibility` categories/types correct? Previously we queried by name... How does it look now when we call `search` function in the SDKs?

src/download/exchangeDownloader.ts
```diff
@@ -228,20 +230,15 @@ export async function searchExchange(
  * @export
  * @param {string} accessToken
  * @param {RestApi} restApi
- * @param {RegExp} [deployment]
  * @returns {Promise<string>} Returned the version string from the instance fetched asset.version value
  */
-export async function getVersionByDeployment(
+export async function getApiVersions(
```
joeluong-sfcc, line 235: It's not clear in the description or the function name that we get the latest version, we should rename this function to something like `getLatestApiVersions` and/or update the JS doc description

joeluong-sfcc (commented):
This is a breaking change correct? Other than the minor comments, this PR LGTM, but could you update the CHANGELOG for this change?

src/download/exchangeDownloader.ts
unandyala: For node sdk, it will be 
`
await download.search(
      'category:Visibility = "External" category:"SDK Type" = "Commerce"'
    );
`

For isomorphic sdk, it will be 
`
await download.search(
      'category:Visibility = "External" category:"SDK Type" = "Commerce"'  category:"SDK Type" = "Isomorphic"'
    );
`

src/download/exchangeDownloader.ts
unandyala: Refactored to put all the new code behind isOas flag

vcua-mobify (dismissed):
Thanks for refactoring this PR to be a non-breaking change! Good idea to use the isOas to feature flag all of the changes.

LGTM

joeluong-sfcc (approved):
LGTM, only thing has the CI always failed since we added OAS support in RAML toolkit, or did it start with this PR? Either way we should resolve the CI so that tests pass

unandyala:
> LGTM, only thing has the CI always failed since we added OAS support in RAML toolkit, or did it start with this PR? Either way we should resolve the CI so that tests pass

this is an existing issue not related to the PR - /bin/sh: 1: oasdiff: not found
Error: oasdiff is not installed. Install oasdiff according to https://github.com/oasdiff/oasdiff#installation

vcua-mobify (approved):
Thanks for fixing the CI issue @unandyala

joeluong-sfcc (approved):

PR #233, opened 2025-06-14
W 18764378/oasdiff on multi vers of api

alexvuong:
This PR fixed the oasdiff command to be compatible with multiple version of the same API. (E.g shopper-basket v2 and v1)
### Problem:
Since we are pulling multiple verserions of the same apis. Some of them will have the same endpoints. When we try to run `oasdiff changlog` on the parent directory (apis vs oldApis). We ran into an error of endpoint duplication.

### Solution
Loop through the each API folder and run oasdiff instead of the parent directory. This will guarantee there is not duplication in each run and concat the results into a txt file.

### How to test drive
- check out code
- npm i
- npm run build
- npm run test
- Observe all the tests are passing

### How to test the changes on isomorphic
- check out this oas branch for isomorphic https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/198
- Change raml-toolkit to locale dependencies
- `yarn ci`
- Make sure your node_modules has the raml-toolkit of your local 
- Run `npm run updateApis` twice to get the current api copied to oldApis folder
- Here two copies will be identical, you won't see much output in the diffApi.txt. Go to oldApis and randomly removed any endpoint from 2 api. E.g oldApis/shopper-basket-oas-1.8.19/shopper-basket-oas-v1-bundled.yaml. Removed end endpoint.  Do the same for shopper-basket-oas-2.x
- Run `yarn run diffApis` and observe the diffApis. You should see output look similiar to this. 
default format output
```
=== Changes in shopper-baskets-oas-1.8.19 ===
1 changes: 0 error, 0 warning, 1 info
info	[endpoint-added] at apis/shopper-baskets-oas-1.8.19/shopper-baskets-oas-v1-bundled.yaml	
	in API POST /organizations/{organizationId}/baskets
		endpoint added



=== Changes in shopper-baskets-oas-2.0.10 ===
1 changes: 0 error, 0 warning, 1 info
info	[endpoint-added] at apis/shopper-baskets-oas-2.0.10/shopper-baskets-oas-v2-bundled.yaml	
	in API DELETE /organizations/{organizationId}/baskets/{basketId}/coupons/{couponItemId}
		endpoint added



======shopper-orders-oas-1.4.8 API is deleted======
======shopper-stores-oas-1.0.15 API is added======
```

- Json output console
```
[
  {
    "directory": "shopper-baskets-oas-1.8.19",
    "changes": [
      {
        "id": "endpoint-added",
        "text": "endpoint added",
        "level": 1,
        "operation": "POST",
        "operationId": "createBasket",
        "path": "/organizations/{organizationId}/baskets",
        "source": "apis/shopper-baskets-oas-1.8.19/shopper-baskets-oas-v1-bundled.yaml",
        "section": "paths"
      }
    ]
  },
  {
    "directory": "shopper-baskets-oas-2.0.10",
    "changes": [
      {
        "id": "endpoint-added",
        "text": "endpoint added",
        "level": 1,
        "operation": "DELETE",
        "operationId": "removeCouponFromBasket",
        "path": "/organizations/{organizationId}/baskets/{basketId}/coupons/{couponItemId}",
        "source": "apis/shopper-baskets-oas-2.0.10/shopper-baskets-oas-v2-bundled.yaml",
        "section": "paths"
      }
    ]
  },
  {
    "directory": "shopper-orders-oas-1.4.8",
    "status": "deleted",
    "message": "shopper-orders-oas-1.4.8 API is deleted"
  },
  {
    "directory": "shopper-stores-oas-1.0.15",
    "status": "added",
    "message": "shopper-stores-oas-1.0.15 API is added"
  }
]

```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,7 +19,7 @@
     "max-lines": [
       "error",
       {
-        "max": 600
+        "max": 700
       }
     ],
     "@typescript-eslint/naming-convention": [
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,7 +19,7 @@
     "max-lines": [
       "error",
       {
-        "max": 700
+        "max": 500
       }
     ],
     "@typescript-eslint/naming-convention": [
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,7 +19,7 @@
     "max-lines": [
       "error",
       {
-        "max": 500
+        "max": 600
       }
     ],
     "@typescript-eslint/naming-convention": [
--- a/src/diff/oasDiff.ts
+++ b/src/diff/oasDiff.ts
@@ -54,8 +54,6 @@ async function findExchangeDirectories(
           subdirectories.push(entryPath);
         }
       }
-      console.log("subdirectories", subdirectories);
-      console.log("hasExchangeJson", hasExchangeJson);
       // If this is a leaf directory and we haven't found exchange.json, that's an error
       if (subdirectories.length === 0 && !hasExchangeJson) {
         throw new Error(
@@ -93,6 +91,36 @@ async function findExchangeDirectories(
   }
 
   return result;
+}
+
+/**
+ * Find YAML files in a directory
+ *
+ * @param dirPath - The directory path to search
+ * @returns Array of YAML file paths
+ */
+async function findYamlFiles(dirPath: string): Promise<string[]> {
+  try {
+    const entries = await fs.readdir(dirPath);
+    const yamlFiles = [];
+
+    for (const entry of entries) {
+      const entryPath = path.join(dirPath, entry);
+      const stat = await fs.stat(entryPath);
+
+      if (
+        stat.isFile() &&
+        (entry.endsWith(".yaml") || entry.endsWith(".yml"))
+      ) {
+        yamlFiles.push(entryPath);
+      }
+    }
+
+    return yamlFiles;
+  } catch (err) {
+    console.warn(`Warning: Could not read directory ${dirPath}:`, err.message);
+    return [];
+  }
 }
 
 /**
@@ -166,7 +194,6 @@ export async function oasDiffChangelog(baseApi: string, newApi: string, flags) {
     console.log("......Starting oasdiff......");
 
     const jsonMode = flags.format === "json" ? "-f json" : "";
-    const directoryMode = flags.dir ? "--composed" : "";
 
     let hasChanges = false;
     let hasErrors = false;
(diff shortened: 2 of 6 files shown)


vcua-mobify (dismissed):
I've verified that these changes work in the node sdk to address the duplicate endpoints errors that appear as a result of having `-internal` and `-public` yaml files in the same location.

I just have a small comment about maybe separating the logic for directory mode but otherwise LGTM.

.eslintrc.json
```diff
@@ -19,7 +19,7 @@
     "max-lines": [
       "error",
       {
-        "max": 500
+        "max": 600
```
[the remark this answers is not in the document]
alexvuong: I disabled this rules for diff files instead

vcua-mobify (dismissed):
Thanks for adjusting @alexvuong !

src/diff/oasDiff.ts
```diff
@@ -4,8 +4,183 @@
  * SPDX-License-Identifier: BSD-3-Clause
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+/* eslint header/header: "off", max-lines:"off" */
```
unandyala, line 7: why do we have to turn off the header rule?

src/diff/oasDiff.ts
```diff
@@ -4,8 +4,183 @@
...
+      for (const entry of entries) {
+        const entryPath = path.join(currentPath, entry);
+        const stat = await fs.stat(entryPath);
+        if (stat.isDirectory()) {
+          subdirectories.push(entryPath);
+        }
+      }
+      // If this is a leaf directory and we haven't found exchange.json, that's an error
+      if (subdirectories.length === 0 && !hasExchangeJson) {
+        throw new Error(
```
unandyala, line 118: I think we can log it as a warning and proceed instead of throwing an error and exiting

src/diff/oasDiff.ts
```diff
@@ -14,71 +189,327 @@ import { execSync } from "child_process";
...
+                if (outputJson?.length > 0) {
+                  directoryChanges.push(...outputJson);
+                }
+              } else {
+                directoryChangesText.push(
+                  `--- Changes in ${baseFileName} ---\n${oasdiffOutput}`
+                );
+              }
+            }
+          } else {
```
unandyala, line 348: nit: Is this case even possible.. where the directory exists but Yaml is missing? May be when download has some issues

src/diff/oasDiff.ts
```diff
@@ -14,71 +189,327 @@ import { execSync } from "child_process";
...
+      continue;
+    }
+
+    // Both directories exist, compare individual yml files in each directory
+    if (baseDir && newDir) {
+      console.log("===================================");
+      console.log(`Processing directory pair: ${dirName}`);
+
+      try {
+        const baseYamlFiles = await findYamlFiles(baseDir.path);
```
unandyala, line 315: Multiple yaml files? Is this because we now have internal and public yaml files?

src/diff/oasDiff.ts
alexvuong: Good point. I think it is stil good to leave it here if download happens to has some issue and dev does not realize it and run oasdiff. The logic in here simply notify which yml files are removed from previous version into output file

unandyala (approved):

vcua-mobify (approved):

PR #235, opened 2025-08-01
Fix: diff command (@W-19076728@)

joeluong-sfcc:
This PR modifies the behavior of the `diff` command, where if the directory name contains the version number, it'll remove the minor and the patch version. For example:

- Before: `shopper-stores-oas-1.0.16`
- After: `shopper-stores-oas-1`

To disable this behavior, a user can pass in the `--disable-normalize-directory-names`  flag.

This is because when we run the `updateApis` command, we append the version number. The current implementation of the `diff` command on directory mode looks for matches in directory names, but since directories have the version number, there will never be a match unless the API version does not change. Example diff file:

```
======assignments-oas-1.0.34 API is deleted======
======campaigns-oas-1.0.36 API is deleted======
======catalogs-oas-1.0.39 API is deleted======
...
======assignments-oas-1.0.35 API is added======
======campaigns-oas-1.0.37 API is added======
======catalogs-oas-1.0.41 API is added======
```

By normalizing the directory names to only include the major version in the directory name, we can start to compare the old API spec with the new API spec. We keep the major version for APIs with multiple versions, like shopper baskets. Example diff file with with updated normalization behavior:

```
=== Changes in shopper-baskets-oas-1 ===
--- Changes in shopper-baskets-oas-v1-public.yaml ---
...

=== Changes in shopper-baskets-oas-2 ===
--- Changes in shopper-baskets-oas-v2-public.yaml ---
...
```

This PR also removes some unnecessary test files that weren't being used.

To test this PR:
1. in RAML toolkit, `git checkout ju/fix-diff-command`
2. `npm run build`
3. Checkout this isomorphic SDK PR: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/208 - `git checkout ju/fix-diff-command`
4. Set up `ANYPOINT_USERNAME` and `ANYPOINT_PASSWORD` env variables via shared password vault
5. In isomorphic SDK: `yarn ci && yarn updateApis`
6. verify in `temp/diffApis.txt` that diff shows up


src/download/exchangeDownloader.ts
```diff
@@ -442,5 +443,5 @@ function getLatestReleaseVersion(versionGroup: {
     if (aMajor !== bMajor) return bMajor - aMajor;
     if (aMinor !== bMinor) return bMinor - aMinor;
     return bPatch - aPatch;
-  })[0].version;
+  })[0]?.version;
```
joeluong-sfcc, line 446: Was seeing an error in isomorphic SDK where we tried to read `version` property from undefined

src/diff/diffCommand.ts
```diff
@@ -70,6 +70,11 @@ Exit statuses:
         "Specifies the API spec of the files being compared. Defaults to RAML. Options are RAML or OAS",
       options: ["raml", "oas"],
     }),
+    "normalize-directory-names": flags.boolean({
+      description:
+        "Normalize directory names by removing minor and patch versions. Example: 'shopper-stores-oas-1.0.16' -> 'shopper-stores-oas-1",
+      default: false,
```
vcua-mobify, line 76: Just wondering; wouldn't the most common use case have us set this to true so that we can compare the difference between two minor versions of the same API? 

Otherwise, the diff will say the old version is removed and the new version is added and leave it at that, no? I don't think this old-version-removed/new-version-added also does a file by file compare?

vcua-mobify (approved):
Aside from a question around whether this should be enabled by default, the changes seem good to me.

Followed the test steps and the resulting diff file was produced

alexvuong (approved):

testResources/setup.ts
```diff
@@ -1,21 +0,0 @@
...
- */
-import sinon from "sinon";
-import { ramlToolLogger } from "../src/common/logger";
-
-let ramlToolLoggerStub: sinon.SinonStub;
-let consoleStub: sinon.SinonStub;
-
-beforeEach(() => {
-  ramlToolLoggerStub = sinon.stub(ramlToolLogger, "info");
-  consoleStub = sinon.stub(console, "log");
```
alexvuong, line 15: This set up is used to silent all console logs in tests. Are you sure you want to remove it?

joeluong-sfcc: Ah, good point, I didn't think about silencing the console logs. I went back to my original approach where I export the `consoleStub`

src/diff/diffCommand.ts
vcua-mobify: Thanks for making this change, @joeluong-sfcc !

alexvuong (approved):

